PR #19, opened 2020-05-12
Simplify cluster spec

vladdy:
This introduces a small set of improvements around the cluster spec and reporting:

1. Topology support was removed as it was found unnecessary.
2. Automated cert generation is now done when no node TLS secret is provided, so the previously used keyword `GENERATED` is also not needed now
3. `COCKROACH_CHANNEL` is set to `kubernetes-operator`
4. `Initialized` condition was renamed to `NotInitialized` to conform the API recommendations that unusual state should have `True` condition.


pkg/actor/deploy.go
```diff
@@ -70,29 +70,16 @@ func (d deploy) Act(ctx context.Context, cluster *resource.Cluster) error {
...
-		if err != nil {
-			return err
-		}
-
-		return nil
-	})
-
+	changed, err = (resource.Reconciler{
+		ManagedResource: r,
+		// Builder:         resource.NewStatefulSetBuilder(cluster, name, nodesNum, join, locality, nodeSelector),
```
joshimhoff, line 75: Remove commented out line?

PR #20, opened 2020-05-17
Refactor resource reconciliation

vladdy:
This refactoring consists of two things:

1. makes resource reconciliation more robust, particularly for cases of different k8s versions which can handle setting default resource property versions differently. This is now handled using https://github.com/banzaicloud/k8s-objectmatcher which keeps the original config as an annotation to the resources and makes comparisons against it

2. reconciliation implementation was simplified (although it is hard to believe in that looking into amount of changes in this PR) and got more tests


vladdy:
@joshimhoff, I wonder if you can find a moment to review this.

pkg/resource/statefulset.go
```diff
@@ -30,23 +31,27 @@ type StatefulSetBuilder struct {
...
 }
 
-func (b StatefulSetBuilder) Build() (runtime.Object, error) {
-	ss := &appsv1.StatefulSet{
-		ObjectMeta: metav1.ObjectMeta{
-			Name: b.StatefulSetName(),
+func (b StatefulSetBuilder) Build(obj runtime.Object) error {
+	ss, ok := obj.(*appsv1.StatefulSet)
+	if !ok {
+		return errors.New("failed to access StatefulSet object")
```
chrisseto, line 37: nit: access is a bit misleading here. Maybe just `failed to cast to StatefulSet`?

PR #21, opened 2020-05-19
Operator deployment

johnrk:
**Adding Operator Deployment file to use Operator on actual K8s Cluster**
This PR includes a new deployment file to test the Operator an actual K8s Cluster.

**Context:**
Previously, we tested the Operator locally. But we need to acceptance test it on actual K8s cluster before sharing it with prospects. This is based on an old deployment file that Vlad shared with me separately a few weeks ago. It points to an image hosted on GCP. That value needs to be replaced by an image that will be hosted on Quay or Docker Hub.

**Testing:**
Pre-requisite: have a kubernetes cluster running.
1. apply CRD
kubectl apply -f config/crd/bases/crdb.cockroachlabs.com_crdbclusters.yaml 
2. apply new Operator deployment file
kubectl apply -f deploy/operator.yaml 
3. initialize a CockroachDB cluster
kubectl apply -f config/examples/example.yaml

**Results:**
1. kubectl get pods
NAME                            READY   STATUS    RESTARTS   AGE
crdb-0                          1/1     Running   0          3h29m
crdb-operator-68d986bdd-889hw   1/1     Running   0          3h31m

2. successfully opened SQL client
kubectl exec -it crdb-0 -- ./cockroach sql --certs-dir=cockroach-certs
it worked.

In addition, I added an example config file that uses "emptyDir: {}"


config/examples/example.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: crdb.cockroachlabs.com/v1alpha1
+kind: CrdbCluster
+metadata:
+  name: crdb
+spec:
+  dataStore:
+    emptyDir: {}
+  tlsEnabled: true
+  nodeTLSSecret: GENERATED
+  image: cockroachdb/cockroach-unstable:v20.1.0
```
joshimhoff, line 10: Could use `cockroachdb/cockroach`, I think.

PR #26, opened 2020-05-31
[WIP] statefulset native scaling

chrisseto:
Wanted to toss this up to get others ideas. The idea is to leverage k8s lifecycle hooks and a custom binary for implement cluster scaling in a K8s "native" way.

@vladdy I'm a bit lost on the best way to add the service accounts, would you mind pointing me in the right direction?

The service account allows the pod to get the statefulset to see if it is restarting or being removed by comparing its index to the Replicas field. If it is being removed, it will invoke `cockroach node decommission`.

I'm also planning to add a prestart hook or update the existing readiness probe to allow minor version patches to be handled by editing the statefulset. Then the operator will just have to handle setting the downgrade cluster setting 🤔


vladdy:
I had to do a better job about announcing it, but I already have this implemented in my local repo the same way as it was done in CC. However, if you want to take on implementing this once more or some other way, I'm fine with that as well as it is a good feature to get familiar with the operator. Let me know if you want to take this over so I could work on other things.

Now regarding your questions:

> Wanted to toss this up to get others ideas. The idea is to leverage k8s lifecycle hooks and a custom binary for implement cluster scaling in a K8s "native" way.

Unfortunately, I'm not fully convinced that the proposed approach is better than the straightforward one. The most apparent difference is that it removes the support of "vanilla" CockroachDB images. Also, it would be required to test scenarios when more than 1 node is removed and added back to ensure there are no corner cases

> I'm a bit lost on the best way to add the service accounts, would you mind pointing me in the right direction?

I would expect `Reconciler` to work for it. If it does not for a some reason, you can always use `client.Create(context.TODO(), serviceAccount)`.

> I'm also planning to add a prestart hook or update the existing readiness probe to allow minor version patches to be handled by editing the statefulset. Then the operator will just have to handle setting the downgrade cluster setting

I'm curious to hear more about this.

chrisseto:
> I had to do a better job about announcing it, but I already have this implemented in my local repo the same way as it was done in CC. 

No worries! 

> However, if you want to take on implementing this once more or some other way, I'm fine with that as well as it is a good feature to get familiar with the operator.

For the moment this is me getting up to speed with the operator and tossing around some ideas.

> Unfortunately, I'm not fully convinced that the proposed approach is better than the straightforward one. The most apparent difference is that it removes the support of "vanilla" CockroachDB images.

I'm split on that. If this method works out well, I would like to pull it into CC to reduce the amount of code that handles updating the statefulset directly. It's a bit easier to share the implementations if it's just an additional binary 🙂.
The biggest downside is there's not a way to gracefully prevent scaling down if replicas can't be relocated.

>  Also, it would be required to test scenarios when more than 1 node is removed and added back to ensure there are no corner cases

I was going to try to use the pre-start hook to run `cockroach node recommission`. It seems that running recommission on new or otherwise active nodes has no ill affects!

> I would expect Reconciler to work for it. 
For whatever reason I kept running into "already exists errors" or if I left in this block:

```go
	if changed {
		log.Info("created/updated service account, stopping request processing")
		CancelLoop(ctx)
		return nil
	}
```

the loop would be canceled but never rerun. I previously dropped this into the deploy actor. The service account would be created but the statefulset was never created 🤔.
I haven't figured out why creating the discover service does not result in the same behavior. Perhaps I'll do some more digging.

> I'm curious to hear more about this.

Part one is adding in the recommission command. I _suspect_ that our code in CC for checking for under replicated ranges, node liveness, and node status could be added to lifecycle hooks to have kubernetes handle all of this itself. 


Again, mostly just trying out some thoughts and looking for feedback. Thank you for taking the time!

vladdy:
> For the moment this is me getting up to speed with the operator and tossing around some ideas.

👍

> I'm split on that. If this method works out well, I would like to pull it into CC to reduce the amount of code that handles updating the statefulset directly. It's a bit easier to share the implementations if it's just an additional binary 🙂.
The biggest downside is there's not a way to gracefully prevent scaling down if replicas can't be relocated.

I definitely see how such hooks can be useful for CC, but I'm worried that it puts specific requirements on the supported images.

> I was going to try to use the pre-start hook to run cockroach node recommission. It seems that running recommission on new or otherwise active nodes has no ill affects!

Yep, I used it in CC.

> the loop would be canceled but never rerun. I previously dropped this into the deploy actor. The service account would be created but the statefulset was never created 🤔.

Those details make it a bit clear on what could be wrong. I would recommend to switch from Cluster roles to regular roles as the operator is assumed to run per namespaces. Also, please ensure you add the account into https://github.com/cockroachdb/cockroach-operator/blob/master/pkg/controller/cluster_controller.go#L102. That ensures that the operator will be aware of any change to the account which means that the reconciliation loop will be started on the creation.

> I suspect that our code in CC for checking for under replicated ranges, node liveness, and node status could be added to lifecycle hooks to have kubernetes handle all of this itself.

Assuming there is a work is being done around proper node restarts and decommissioning in 20.1 (it probably has been even completed, I have not checked), I'm not sure if we need to duplicate the logic.

vladdy:
@chrisseto, just checking if you still intend to work on this.

chrislovecnm:
@chrisseto hope you are having a great day!  Are you planning on doing more work on this PR?

cc: @johnrk

chrisseto:
Sorry for the late response! I don't plan to pursue this PR any further.

chrislovecnm:
@chrisseto no worries!  Thanks for the response

PR #27, opened 2020-06-01
Version upgrade

vladdy:
Import upgrade code from my original implementation and add e2e tests for it.


joshimhoff:
Will review today, sorry for the slowness.

vladdy:
@joshimhoff not a problem! Take your time!

e2e/e2e_test.go
```diff
@@ -0,0 +1,206 @@
...
+		{
+			name: "creates a 1-node secure cluster",
+			test: func(t *testing.T) {
+				require.NoError(t, sb.Create(builder.Cr()))
+
+				requireClusterToBeReadyEventually(t, sb, builder)
+			},
+		},
+		{
+			name: "upgrades the cluster to the next minor version",
```
joshimhoff, line 148: "next minor version" -> "next patch version".

pkg/actor/upgrade.go
```diff
@@ -0,0 +1,192 @@
...
+	}
+
+	return rows[1][0], nil
+}
+
+func (up *upgrade) showClusterSetting(cluster *resource.Cluster, setting string) ([][]string, error) {
+	cmd := []string{
+		"/bin/bash",
+		"-c",
+		"/cockroach/cockroach sql " + cluster.SecureMode() +
```
joshimhoff, line 148: How does this work in secure mode? Are certs sitting in the operator container? Which certs & how does this happen?

pkg/actor/upgrade.go
```diff
@@ -0,0 +1,192 @@
...
+
+	if rows[0][0] != setting {
+		return "", fmt.Errorf("expected column named \"%s\", got: %s", setting, rows[0][0])
+	}
+
+	return rows[1][0], nil
+}
+
+func (up *upgrade) showClusterSetting(cluster *resource.Cluster, setting string) ([][]string, error) {
+	cmd := []string{
```
joshimhoff, line 145: What do you think about establishing a SQL connection from the operator instead of shelling out? I'd expect that approach to be a bit more maintainable. I think the value of doing it thru a proper SQL connections depends a bit on often we think the k8s operator will need to make SQL calls.

pkg/actor/upgrade.go
```diff
@@ -0,0 +1,192 @@
...
+		return NotReadyErr{Err: errors.New("statefulset is updating, waiting for the update to finish")}
+	}
+
+	// Get version from spec
+	specStr := getVersionFromImage(cluster.Spec().Image)
+	if specStr == "" {
+		return fmt.Errorf("Unknown CockroachDB version in spec: %s", cluster.Spec().Image)
+	}
+
+	specVer, err := semver.NewVersion(specStr)
```
joshimhoff, line 63: We are actually considering moving away from deriving versions from a docker tag. An alternative approach we are considering is calling `cockroach version` to get the version. Anyway, this looks good for now, just sharing that we are discussing this with you.

pkg/actor/upgrade.go
```diff
@@ -0,0 +1,192 @@
+package actor
+
+import (
+	"context"
+	"encoding/csv"
+	"fmt"
+	"github.com/Masterminds/semver/v3"
```
joshimhoff, line 7: OH NICE!

e2e/e2e_test.go
```diff
@@ -0,0 +1,206 @@
...
```
joshimhoff, line 191: Can you also test that preserve downgrade option is being set here but not when doing a patch version? I understand that if we fail to do this, the cluster won't become ready (at least that's what I think will happen), but still seems prudent to test.

e2e/e2e_test.go
```diff
@@ -0,0 +1,206 @@
...
+
+	testLog := zapr.NewLogger(zaptest.NewLogger(t))
+
+	actor.Log = testLog
+
+	sb := testenv.NewDiffingSandbox(t, env)
+	sb.StartManager(t, controller.InitClusterReconcilerWithLogger(testLog))
+
+	builder := testutil.NewBuilder("crdb").WithNodeCount(1).WithTLS().
+		WithImage("cockroachdb/cockroach:v19.2.6").
```
joshimhoff, line 178: Can you test an upgrade from `v19.1.4` to `v19.2.1` or something also? Is a bit different in terms of the change in version tag but also a major release according to our definitions at CRL.

e2e/e2e_test.go
```diff
@@ -0,0 +1,206 @@
+package e2e
```
joshimhoff, line 1: How would I trigger an update? Update README or something?

pkg/actor/upgrade.go
```diff
@@ -0,0 +1,192 @@
...
+	if err != nil {
+		return errors.Wrapf(err, "failed to parse spec image version: %s", specStr)
+	}
+
+	clusterStr, err := up.getClusterSetting(cluster, "version")
+	if err != nil {
+		return errors.Wrapf(err, "failed to fetch cluster version")
+	}
+
+	contraintStr := fmt.Sprintf("^%s", clusterStr)
```
joshimhoff, line 73: Can you add a constant explaining how this works and maybe link docs?

pkg/actor/upgrade.go
```diff
@@ -0,0 +1,192 @@
...
+	action
+
+	config *rest.Config
+}
+
+func (up *upgrade) Handles(conds []api.ClusterCondition) bool {
+	return condition.False(api.NotInitializedCondition, conds)
+}
+
+func (up *upgrade) Act(ctx context.Context, cluster *resource.Cluster) error {
```
joshimhoff, line 38: Finalizing the release is done outside the operator? Any plans to change this?

pkg/actor/upgrade.go
vladdy: Yep, the certificates exist in all pods in secure mode.

pkg/actor/upgrade.go
vladdy: I'm not against that and my only concern would be having two ways to get data/manage cluster. We will need to clarify when use what in such case. The current way to send the query relies on the existing implementation to get node status or initialize the cluster.

pkg/actor/upgrade.go
vladdy: Noted and sounds good to me. It won't be hard to switch to this. Just create an issue or PR when the decision is made.

e2e/e2e_test.go
vladdy: Sorry, did you mean assertions? If yes, I actually think we need to keep e2e tests clean out of implementation details. However, I think we need some kind of fake implementation of interaction with crdb for functional tests where we would have check for that. I or somebody could look into that later.

e2e/e2e_test.go
vladdy: I actually thought about having 19.2 as the minimal supported version by the operator, but if you really think we need such test, I could add it.

e2e/e2e_test.go
vladdy: Docs are incoming =)

pkg/actor/upgrade.go
vladdy: Yes, currently, I think it is better to leave to some external party (a human or an automation) to finalize the update. Later once we have more feedback, we can revise it. The main problem here as it appears to me is how the operator can verify that it is fine to finalize as it does not have feedback from the applications using the database.

e2e/e2e_test.go
vladdy: I've added an additional e2e test.

PR #67, opened 2020-07-13
Updating Test Image to have gcloud

chrislovecnm:
Adding gcloud to the test image


chrislovecnm:
Closing this for now as it will not address the problem fully

Issue #79, opened 2020-07-14
PodDisruptionBudget


chrislovecnm:
Add API values to control PDB, and implement PDB.

```
maxUnavailable: 1
```

Probably need to default to a value that will keep quorum.

chrislovecnm:
@johnrk requirements questions.  We do not allow the user to set `minAvailable`, in the helm chart, instead of `maxUnavailable` in the PDB.  I have implemented the capability for a user to set both.  Thoughts?

If the size of the database is just a single pod we do not create the PDB. 

Open question.  Do we allow the user to not have a PDB at all?  It is a valid use case, but I would not recommend the edge case, but you never know with users :)

chrisseto:
> If the size of the database is just a single pod we do not create the PDB.

We should probably note somewhere that single node clusters are not suitable for production use?

> Do we allow the user to not have a PDB at all? It is a valid use case,

What cases do you have in mind besides a single node cluster?

chrislovecnm:
@chrisseto 

> What cases do you have in mind besides a single node cluster?

Some users do not like PDBs, it is personal preference tbh.

chrislovecnm:
> We should probably note somewhere that single node clusters are not suitable for production use?

Is this not documented in the DB docs?

vladdy:
so if some users do not like it, why do we need to support it? There is nothing what prevents the user from creating a CR with the budget they need along with cockroachdb cluster. 

The only pro of supporting PDBs in the operator I see is it allows to remove the logic of ensuring that the db cluster has enough of healthy nodes out of the operator. However, it is unclear how this is going to play with auto-scaling. Hopefully, these things won't intersect =).

chrislovecnm:
@johnrk we want the parity with the helm chart for PDBs?  I am adding a feature and will have parity as well.  Do we want to allow the user to disable the PDB?

chrislovecnm:
@johnrk the initial code is in, but I have testing that I am planning.

chrislovecnm:
Closing this, as we have other issues open for testing

johnrk:
@chrislovecnm , I'm just now catching up with this issue. Let's go for parity with the Helm Chart. What initial code is in? What did we land on here?

keith-mcclellan:
@chrislovecnm what's the status on this? we should be defaulting this to be the number of cockroachdb nodes -1 by default since the default RF = 3

chrislovecnm:
Fixed

PR #82, opened 2020-07-15
Implementing support for PDB

chrislovecnm:
This implements https://github.com/cockroachdb/cockroach-operator/issues/79

@chrisseto when you merge if you could not squash this time, please.

This PR includes

1. New API fields for setting a Max or a Min PBD
2. Functionality that allows the controller to CRUD PBDs
3. Unit tests
4. Modified an integration test to allow for this support
5. made a small Makefile change, which is not related to #79 here https://github.com/cockroachdb/cockroach-operator/pull/82/commits/f9ab7af223aa564d1dde1a574c75f8bbdbd8746f


chrislovecnm:
@vladdy and @chrisseto PTAL

@johnrk I have comments on https://github.com/cockroachdb/cockroach-operator/issues/79 about this support.

api/v1alpha1/cluster_types.go
```diff
@@ -22,8 +22,10 @@ import (
 )
 
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+// Important: Run "make" to regenerate code after modifying this file
```
chrisseto, line 25: Maybe note the specific make task/target?

api/v1alpha1/cluster_types.go
```diff
@@ -22,8 +22,10 @@ import (
...
 
-// CrdblusterSpec defines the desired state of Cluster
+// CrdblusterSpec defines the desired state of CockRoachDB Cluster
```
chrisseto, line 27: nit:

```suggestion
// CrdblusterSpec defines the desired state of a CockroachDB Cluster
```

api/v1alpha1/cluster_types.go
```diff
@@ -50,6 +52,13 @@ type CrdbClusterSpec struct {
 	// (Optional) The secret with a certificate and a private key for root database user
 	// Default: ""
 	ClientTLSSecret string `json:"clientTLSSecret,omitempty"`
+	// TODO document on how you need to set the max or the min
+	// The maximum number of pods that can be unavailable during a rolling update.
+	// This number is set in the PodDistruptionBudget and defaults to 1.
+	MaxUnavailable *int32 `json:"maxUnavailable,omitempty"`
+	// The min number of pods that can be unavailable during a rolling update.
```
chrisseto, line 59: The min number of pods that will be available during a rolling update?

The description and variable name seem to diverge.

pkg/testutil/builder.go
```diff
@@ -108,6 +108,11 @@ func (b ClusterBuilder) WithImage(image string) ClusterBuilder {
 	return b
 }
 
+func (b ClusterBuilder) WithMaxUnavaible(max *int32) ClusterBuilder {
```
chrisseto, line 111:
```suggestion
func (b ClusterBuilder) WithMaxUnavailable(max *int32) ClusterBuilder {
```

pkg/resource/pod_distruption_budget.go
```diff
@@ -0,0 +1,94 @@
...
+	if pdb.ObjectMeta.Labels == nil {
+		pdb.ObjectMeta.Labels = map[string]string{}
+	}
+
+	// TODO test sevice name
+	// Can we use the common label selector?
+	// commonLabels := labels.Common(cluster.Cr())
+	//  selector: commonLabels.Selector(),
+
+	labelSelector, err := metav1.ParseToLabelSelector("app=" + b.DiscoveryServiceName())
```
chrisseto, line 49: Food for thought comment: Should we be looking move to `app.kubernetes.io/name` rather than `app`?

pkg/resource/pod_distruption_budget.go
```diff
@@ -0,0 +1,94 @@
...
+	if pdb.ObjectMeta.Name == "" {
+		pdb.ObjectMeta.Name = b.DiscoveryServiceName()
+	}
+
+	if pdb.ObjectMeta.Labels == nil {
+		pdb.ObjectMeta.Labels = map[string]string{}
+	}
+
+	// TODO test sevice name
+	// Can we use the common label selector?
```
chrisseto, line 45: Seems desirable as running two crdb clusters in the same CRDB cluster couldn't currently have multiple PDBs if I'm reading this correctly?

api/v1alpha1/cluster_types.go
chrislovecnm: It is an automatically generated comment from kubebuilder.  I have a note to catch this in CI, which I think is the best solution.  I think I will pull the comment.

api/v1alpha1/cluster_types.go
chrislovecnm: It is a typo :)

pkg/resource/pod_distruption_budget.go
chrislovecnm: What do you mean?  There is a func that provides a common label, and I think I can reuse that.  I need to poke around a bit more to see how to convert a map into a label.

pkg/resource/pod_distruption_budget.go
chrislovecnm: I need to look at our labels.  We need to be looking for a label on the service that is super unique.

pkg/resource/pod_distruption_budget.go
chrislovecnm:
```yaml
   app.kubernetes.io/instance: crdb
```

But to be honest we should have a

```yaml
  app.cockroachlabs.com/instance: crdb
```

Or something unique like that. Yes looking at it now the implementation needs to change, but we also might want better labels.  Thoughts?

chrisseto: It looks like the PDB selector is just `app=cockroachdb`. Won't that conflict if I create to cluster with different PDB configurations?

pkg/resource/pod_distruption_budget.go
chrislovecnm: I'm thinking I want to match the service name since that is guaranteed to be unique.

On the topic of labels.  We have this:

```yaml
  app.kubernetes.io/instance: crdb
```

We may want to change it to:

```yaml
  app.somecockroachurl/instance:crdb
```

Or add a label like the above.

vladdy: Probably something from `pkg/labels/label.go` could be helpful here.

chrislovecnm: See https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/

vladdy:
Just as a side comment, I think squashing commits when merging is better so that there is one batch of changes to revert if it is needed.

chrislovecnm:
@vladdy I am used to Kubernetes project where you don't squash, but no big deal.  If we do squash then I need to pull out the Makefile commit.

api/v1alpha1/cluster_defaults_test.go
```diff
@@ -18,19 +18,21 @@ package v1alpha1
...
+
 	"github.com/google/go-cmp/cmp"
 	"github.com/stretchr/testify/assert"
-	"testing"
 )
 
 func TestSetClusterSpecDefaults(t *testing.T) {
 	s := &CrdbClusterSpec{}
-
+	max := int32(1)
```
vladdy, line 29: nit: probably this variable could be better named as it is unclear what max is it of.

pkg/resource/pod_distruption_budget.go
chrislovecnm: @vladdy yah I think that is what is being used, but for some reason, we are missing some of the labels.  What are your thoughts on adding a custom label?  Or should I just match the static labels?  For instance, the version label is not static.

pkg/resource/pod_distruption_budget.go
```diff
@@ -0,0 +1,94 @@
...
+	// TODO check to see if both are not nil??
+
+	if b.Cluster.cr.Spec.MinAvailable != nil {
+		min := b.Cluster.cr.Spec.MinAvailable
+		minIS := intstr.FromInt(int(*min))
+		pdb.Spec = policy.PodDisruptionBudgetSpec{
+			MinAvailable: &minIS,
+			Selector:     labelSelector,
+		}
+	} else if b.Cluster.cr.Spec.MaxUnavailable != nil {
```
vladdy, line 63: can this be just another `if` as those parameters appear to be independent.

pkg/resource/pod_distruption_budget.go
chrislovecnm: @chrisseto I got yah now.  That is bad code.  We need to decide which labels I am matching, which is the discussion above.

pkg/controller/cluster_controller.go
```diff
@@ -18,16 +18,18 @@ package controller
...
 	"context"
+	"time"
+
 	api "github.com/cockroachdb/cockroach-operator/api/v1alpha1"
 	"github.com/cockroachdb/cockroach-operator/pkg/actor"
 	"github.com/cockroachdb/cockroach-operator/pkg/resource"
 	"github.com/go-logr/logr"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	policy "k8s.io/api/policy/v1beta1"
```
vladdy, line 29: it would be helpful to either rename the above includes to drop `v1` or use the same scheme.

pkg/resource/pod_distruption_budget.go
```diff
@@ -0,0 +1,94 @@
...
+func (b PdbBuilder) Build(obj runtime.Object) error {
+	pdb, ok := obj.(*policy.PodDisruptionBudget)
+	if !ok {
+		return errors.New("failed to cast to PDB object")
+	}
+
+	// if we only have one Node we cannot have a PDB
+	// TODO we need to validate this in the CRD API
+	if b.Spec().Nodes == 1 {
+		return nil
```
vladdy, line 32: This will never be called as the actor always checks for `> 1`

pkg/resource/pod_distruption_budget.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	// TODO fix or should we use this?
+	if pdb.ObjectMeta.Name == "" {
```
vladdy, line 36: IMO, we should not use this as well as 'if' below as this is the reconciliation loop.

chrisseto:
Kubernetes doesn't squash? That seems unmanageable 😬 

We try to stick to atomic commits on CC and CRDB.

pkg/resource/pod_distruption_budget.go
```diff
@@ -0,0 +1,94 @@
...
+	labelSelector, err := metav1.ParseToLabelSelector("app=" + b.DiscoveryServiceName())
+	if err != nil {
+		return fmt.Errorf("unexpected error parsing label: %v", err)
+	}
+
+	// TODO check to see if both are not nil??
+
+	if b.Cluster.cr.Spec.MinAvailable != nil {
+		min := b.Cluster.cr.Spec.MinAvailable
+		minIS := intstr.FromInt(int(*min))
```
vladdy, line 58: I'm not sure what "IS" means in this variable and wonder if it can be dropped and `*b.Cluster.cr.Spec.MinAvailable` can be just used directly in the conversion.

As another thing, can 'intstr.IntOrString' be used in the cluster spec, so that we can avoid the conversion?

pkg/resource/pod_distruption_budget.go
```diff
@@ -0,0 +1,94 @@
...
+	return nil
+}
+
+func (b PdbBuilder) Placeholder() runtime.Object {
+
+	// if we only have one Node we cannot have a PDB
+	// TODO we need to validate this in the CRD API
+	if b.Spec().Nodes == 1 {
+		return nil
+	}
```
vladdy, line 87: This is unnecessary, also the code which can call Placeholder always expects an object.

pkg/resource/pod_distruption_budget_test.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/google/go-cmp/cmp"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	policy "k8s.io/api/policy/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+)
+
+func TestPDBBuilder(t *testing.T) {
+
```
vladdy, line 20: nit: should this blank line be removed?

pkg/resource/pod_distruption_budget_test.go
```diff
@@ -0,0 +1,71 @@
...
+	var MaxInt int32 = 1
```
vladdy, line 21: `maxUnavailable`? Otherwise I'm expecting some limit on the integer, although it can be my personal perception.

pkg/resource/pod_distruption_budget_test.go
```diff
@@ -0,0 +1,71 @@
...
+	cluster := testutil.NewBuilder("test-cluster").Namespaced("test-ns").WithMaxUnavaible(&MaxInt)
+	commonLabels := labels.Common(cluster.Cr())
+
+	labelSelector, err := metav1.ParseToLabelSelector("app=" + cluster.Cr().Name)
+	if err != nil {
+		t.Errorf("unexpected error parsing label: %v", err)
+	}
+
+	max := intstr.FromInt(1)
```
vladdy, line 30: I would suggest to consider a better name here as well as max is also kind of unclear.

vladdy:
> I am used to Kubernetes project where you don't squash, but no big deal. If we do squash then I need to pull out the Makefile commit.

I don't think it is a big deal to pull it out. Or you can squash only related commits before this is ready for the final merge and @chrisseto won't squash it.

pkg/resource/pod_distruption_budget.go
vladdy: I really don't see why we need a custom label if k8s already recommends common ones. Also, if I remember it right this selector should also match the one used in the StatefulSet, so it would be better just to use the same way to specify the selector as it is done in StatefulSet reconciler.

vladdy:
Btw, how is this going to work for the case of 3 nodes -> 1 nodes downscale if it is going to be supported soon (Chris should know better about that)

chrislovecnm:
It needs to be removed it they downsize it to one node.

Let me fix that

vladdy:
> Let me fix that

Wait for @chrisseto to respond before proceeding. It was not that trivial to downscale to 1 node in the past so unless it has been fixed, it is better to declare it as unsupported operation.

chrislovecnm:
Then that in itself sounds like another can of worms

pkg/controller/cluster_controller.go
chrislovecnm: I'm not making those changes in this PR, since it does not deal with this PR, if you want to open an issue that would be amazing.  I think those changes need to happen across multiple files.

api/v1alpha1/cluster_defaults_test.go
chrislovecnm: Updated

pkg/resource/pod_distruption_budget.go
chrislovecnm: I reworked the if statements.  PTAL

pkg/resource/pod_distruption_budget.go
chrislovecnm: Removed

pkg/resource/pod_distruption_budget.go
chrislovecnm: Please explain more

pkg/resource/pod_distruption_budget.go
chrislovecnm: I can check, I am uncertain

pkg/resource/pod_distruption_budget.go
chrislovecnm: What do I put for a code comment on the func?  I updated the code per your comment.

pkg/resource/pod_distruption_budget.go
chrislovecnm: The updated code is pushed.  I am still doing some testing, but PTAL

pkg/resource/pod_distruption_budget.go
chrislovecnm: All of this is updated now.  Still testing, but PTAL

pkg/resource/pod_distruption_budget.go
chrislovecnm: It is `IntOrString` hence `IS`.  Yes, we can use `IntOrString` here, but then I recommend that we convert all of our ints to `IntOrString` in the API.  I would prefer to not have one value an `int32` and another an `IntOrString`.  So either we change all the ints in the API or leave them all the same.  Thoughts??

pkg/resource/pod_distruption_budget.go
chrislovecnm: @vladdy need feedback here please

pkg/resource/pod_distruption_budget_test.go
chrislovecnm: removed

pkg/resource/pod_distruption_budget_test.go
chrislovecnm: changed

pkg/resource/pod_distruption_budget_test.go
chrislovecnm: changed

chrislovecnm:
@chrisseto and @vladdy PTAL.  I have more testing to do, but I think we are getting close to code complete.

Makefile
```diff
@@ -12,15 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Make will use bash instead of sh
```
chrislovecnm, line 15: I going to back out the Makefile changes and work on this in a separate PR

hack/create-gke-cluster.sh
```diff
@@ -34,7 +34,7 @@ source "$ROOT"/common.sh
 # shellcheck disable=SC1090
 source "$ROOT"/functions.sh
 
-# enable googleapi for container registry on the project
+enable googleapi for container registry on the project
```
chrisseto, line 37: Should this still be a comment?

pkg/resource/pod_distruption_budget.go
```diff
@@ -0,0 +1,70 @@
...
+		maxUnavailable := b.Cluster.cr.Spec.MaxUnavailable
+		maxUnavailableIS := intstr.FromInt(int(*maxUnavailable))
+		pdb.Spec.MaxUnavailable = &maxUnavailableIS
+	}
+
+	return nil
+}
+
+// TODO - what does this do???
+func (b PdbBuilder) Placeholder() runtime.Object {
```
chrisseto, line 64: I think this generates the default/minimum `runtime.Object` for this builder.
If I understand correctly, the labels would be set to an empty map here?
I defer to @vladdy.

hack/create-gke-cluster.sh
chrislovecnm: Nice catch. I thought I backed out those changes completely

chrislovecnm: Fixed

chrislovecnm:
@vladdy please feel free to push another PR on this code.  I created two new issues with the open questions we have.  Also opening an issue on e2e testing this.  Working with you on how to do that would be very helpful.  As always I appreciate your input!

PR #85, opened 2020-07-15
Work on linting fixes

chrislovecnm:
Work getting `make lint` happy and working.

I did not fix all the problems with spellcheck since some are nontrial to fix, and not critical to the operator.  I disabled a couple of checks.  The Dockerfiles are not linted yet and I will add and issue.


chrislovecnm:
cc @chrisseto PTAL

PR #86, opened 2020-07-16
updated README - additional comments, deleting pv and pvc

johnrk:
For first time users, it's not intuitive that persistent volumes and persistent volume claims are not deleted when the CockroachDB cluster is deleted from Kubernetes. This additional comment clarifies how to manually remove persistent volumes and persistent volume claims.

cc @chrislovecnm

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,8 +24,7 @@ export CLUSTER=test
 ./hack/push-operator-gcr.sh
 
 # apply the operator
-# replace the -i argument with the image name
-./hack/apply-operator.sh -i myimagename:2342 -c $CLUSTER
+./hack/apply-operator.sh -c $CLUSTER
 
 # validate the the operator is running
 alias kubectl=k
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -25,6 +25,7 @@ export CLUSTER=test
 
 # apply the operator
 # replace the -i argument with the image name
+# If you've already specified which image to use with kustomize, you don't need to specify image below.
```
chrislovecnm, line 28: We can actually just remove the `-i`.  No need to mention it

README.md
```diff
@@ -45,6 +46,9 @@ Clean up the cluster
 # delete the operator
 ./hack/delete-operator.sh -c $CLUSTER
 
+# If you're still using the gke cluster, you can delete persistent volumes and persistent volume claims. It is not recommended to do this in production.
+kubectl delete pv,pvc --all
```
vladdy, line 50: This can delete other existing claims or PVs which still can be dangerous. Probably, it would be enough to mention that we don't do the removal as it is the default in Kubernetes.

PR #92, opened 2020-07-21
Makefile Improvements

chrislovecnm:
- running bash as the default shell, so I had to add bash to update test-runner container
- added variable for the e2e tests, that can allow a single test to run
- found a couple of listing errors


chrislovecnm:
@chrisseto PTAL

PR #96, opened 2020-07-21
Updating Go Imports

chrislovecnm:
- using goimport binary to update go files
- added a python script to update go files
- adding Makefile target


chrislovecnm:
cc @chrisseto

PR #97, opened 2020-07-21
Missed headers on go files

chrislovecnm:
And this is why we need CI.  I missed headers on two new
go files.


chrislovecnm:
cc @chrisseto

PR #107, opened 2020-07-23
Stop auto-mounting SA Token

### LINKED ISSUE #66: Pod security with manifest that is deployed
I need to look at the Pod manifest that we are deploying with the operator.  I think it still has the auth token mounted.  I should also double-check the container security as well.  I might be wrong on the auth token, but we may need to remove it.

chrislovecnm:
By default Kubernetes mounts the default API authentication
token to a Pod.  This PR adds 'automountServiceAccountToken: false'
to the StatefulSet definition for a managed database.

Closes #66

### DIFF
--- a/e2e/testdata/TestCreatesInsecureCluster/creates_3-node_insecure_cluster.golden
+++ b/e2e/testdata/TestCreatesInsecureCluster/creates_3-node_insecure_cluster.golden
@@ -22,6 +22,7 @@ metadata:
     statefulset.kubernetes.io/pod-name: crdb-0
   name: crdb-0
 spec:
+  automountServiceAccountToken: false
   containers:
   - args:
     - shell
@@ -68,9 +69,6 @@ spec:
     volumeMounts:
     - mountPath: /cockroach/cockroach-data/
       name: datadir
-    - mountPath: /var/run/secrets/kubernetes.io/serviceaccount
-      name: default-token-0
-      readOnly: true
   dnsPolicy: ClusterFirst
   enableServiceLinks: true
   priority: 0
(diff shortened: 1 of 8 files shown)


chrislovecnm:
The e2e test is currently failing and I am uncertain why.  Here is the error message:

```
=== CONT  TestCreatesSecureClusterWithGeneratedCert/creates_1-node_secure_cluster
    assert.go:74: unexpected result (-want +got):
          strings.Join({
          	... // 230 identical lines
          	"        app.kubernetes.io/name: cockroachdb",
          	"    spec:",
        + 	"      automountServiceAccountToken: false",
          	"      containers:",
          	"      - args:",
          	... // 80 identical lines
          }, "\n")
--- FAIL: TestCreatesSecureClusterWithGeneratedCert (52.04s)
    --- FAIL: TestCreatesSecureClusterWithGeneratedCert/creates_1-node_secure_cluster (50.18s)
FAIL
FAIL	github.com/cockroachdb/cockroach-operator/e2e	59.852s
FAIL
```

Steps to recreate. 

1. have a k8s cluster, you can run `hack/create-gke-cluster.sh -c test`
2. run `TEST_ARGS="-run TestCreatesSecureClusterWithGeneratedCert" make e2e-test` as this will only run `TestCreatesSecureClusterWithGeneratedCert`.

chrislovecnm:
@chrisseto PTAL

@johnrk we can demo this on Monday.

justinknoepfler:
There should be a way to override this setting. Service account tokens are a very common mechanism for authenticating with credential / config providers. For example, our production services rely on service account token injection to authenticate with Vault, and we use vault to manage AWS credentials. This change makes it almost impossible to securely manage implicit AWS credentials (for things like s3 export) without resorting to machine creds.

There currently isn't any way to mount additional volumes on the pod, nor supply an initContainer to pull creds from a secure location. 

Curious what the work around is for CRDB pods that need access to their service account token.

PR #109, opened 2020-07-23
Pod image API Values

chrislovecnm:
Adding the capability to add:

- pullPolicy for image
- name of the secret that will be used for ImagePullSecret


deploy/kustomization.yaml
```diff
@@ -19,4 +19,4 @@ kind: Kustomization
 images:
```
chrislovecnm, line 19: need to remove this commit.  I have to fix the way we use kustomize so that that this file stops getting modified.

api/v1alpha1/pod_container_types.go
```diff
@@ -0,0 +1,34 @@
...
+
+import v1 "k8s.io/api/core/v1"
+
+// +kubebuilder:object:generate=true
+
+// PodImage represents the image information for a container that is used
+// to build the StatefulSet.
+type PodImage struct {
+	// (Required) Container image with supported CockroachDB version
+	Name string `json:"name,omitempty"`
```
chrisseto, line 27: More of a musing than a comment: Would it be worth while to split `Name` into image and tag? Image could default to `cockroachdb/cockroach`.

pkg/resource/statefulset.go
```diff
@@ -157,14 +157,27 @@ func (b StatefulSetBuilder) makePodTemplate() corev1.PodTemplateSpec {
...
+
+	secret := b.Spec().Image.PullSecret
+	if secret != nil {
+		local := corev1.LocalObjectReference{
+			Name: *secret,
+		}
+
+		var imagePullSecrets []corev1.LocalObjectReference
+		imagePullSecrets[0] = local
+		pod.Spec.ImagePullSecrets = imagePullSecrets
```
chrisseto, line 169: won't this result in a nil reference as slices default to nil?

Would this be considered more idiomatic?
```suggestion
		pod.Spec.ImagePullSecrets = []corev1.LocalObjectReference{local}
```

api/v1alpha1/pod_container_types.go
chrislovecnm: What is the value 🤷‍♂️

pkg/resource/statefulset.go
chrislovecnm: Yes, and I can never remember the damn syntax 😂

api/v1alpha1/pod_container_types.go
chrislovecnm: Also, how do we support digests then?  For example:

```
docker pull frolvlad/alpine-miniconda3:python3.7@sha256:9bc9c096713a6e47ca1b4a0d354ea3f2a1f67669c9a2456352d28481a6ce2fbe
```

There are multiple formats for "tagging", and I would rather just have the user provide the entire string.  Easy :)

chrisseto: Ah, good point! I didn't think of that.

pkg/resource/statefulset.go
```diff
@@ -157,14 +157,27 @@ func (b StatefulSetBuilder) makePodTemplate() corev1.PodTemplateSpec {
...
 		},
 	}
+
+	secret := b.Spec().Image.PullSecret
+	if secret != nil {
+		local := corev1.LocalObjectReference{
+			Name: *secret,
+		}
+
+		var imagePullSecrets []corev1.LocalObjectReference
```
chrisseto, line 167: Tagging the slice issue again

pkg/resource/statefulset.go
```diff
@@ -164,9 +164,7 @@ func (b StatefulSetBuilder) makePodTemplate() corev1.PodTemplateSpec {
 			Name: *secret,
 		}
 
-		var imagePullSecrets []corev1.LocalObjectReference
-		imagePullSecrets[0] = local
-		pod.Spec.ImagePullSecrets = imagePullSecrets
+		pod.Spec.ImagePullSecrets = []corev1.LocalObjectReference{local}
```
chrislovecnm, line 167: @chrisseto I think this is what you wanted.

chrislovecnm:
@chrisseto I am failing a couple of e2e tests because of test data issues. Fixing them now.

PR #123, opened 2020-08-05
Initial set of code comments

### LINKED ISSUE #118: Code Documentation spike
@vladdy has offered to update some of the code documentation.  This issue is the initial start of adding documentation.

- [ ] https://github.com/cockroachdb/cockroach-operator/blob/master/pkg/resource/discovery_service.go - specifically the placeholder func
- [ ] https://github.com/cockroachdb/cockroach-operator/blob/master/pkg/actor/deploy.go - specifically how these actors are called.
- [ ] https://github.com/cockroachdb/cockroach-operator/blob/master/pkg/controller/cluster_controller.go - this is the entry code that I believe is quite important to understand

I will let you decide if we should add documentation here https://github.com/cockroachdb/cockroach-operator/blob/master/cmd/cockroach-operator/main.go#L64 - it is the main entry for setting up the controller, and I am still learning how this works conceptually.  Maybe we can reference some of there documentation?

vladdy:
Code commenting for the most important logic. Let me know if something needs to be expanded.

Closes #118

### DIFF
--- a/pkg/controller/cluster_controller.go
+++ b/pkg/controller/cluster_controller.go
@@ -58,7 +58,16 @@ type ClusterReconciler struct {
 // +kubebuilder:rbac:groups=policy,resources=poddisruptionbudgets,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=policy,resources=poddisruptionbudgets/status,verbs=get
 
+// Reconcile is the reconciliation loop entry point for cluster CRDs.  It fetches the current cluster resources
+// and uses its state to interact with the world via a set of actions implemented by `Actor`s
+// (i.e. init cluster, create a statefulset).
+// Each action can result in:
+//   - a short requeue (5 seconds)
+//   - a long requeue (5 minutes)
+//   - cancel the loop and wait for another event
+//   - if no other errors occurred continue to the next action
 func (r *ClusterReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
+	// Ensure the loop does not take longer than 20 seconds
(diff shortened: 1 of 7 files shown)


vladdy:
cc @chrislovecnm @chrisseto as I'm unable to update the reviewers. (can something be done about that?)

pkg/controller/cluster_controller.go
```diff
@@ -58,7 +58,15 @@ type ClusterReconciler struct {
 // +kubebuilder:rbac:groups=policy,resources=poddisruptionbudgets,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=policy,resources=poddisruptionbudgets/status,verbs=get
 
+// Reconcile is the reconciliation loop entry point for cluster CRDs. It fetches the current cluster resources
+// and uses its state to interact with the world via a set of actions (i.e. init cluster, create a statefulset).
```
chrisseto, line 62: nit: set of `Actors`?

vladdy: Yeah, naming is hard. So here is the thing. I called the field `Actions` in the controller struct. Those are assumed to be implemented by `Actors`. I used `actions` here as it is how it is referred in the current package. I've also extended the comment adding that detail.

Let me know if you still think it would be more clear.

vladdy:
@chrisseto RFAL

chrisseto:
Thanks @vladdy. Just checking permissions, you should be able to merge with the an approved review?

vladdy:
@chrisseto,  thanks, but unfortunately, I still don't see the Merge button :(

PR #143, opened 2020-08-21
Changing e2e testing to use kind

chrislovecnm:
We previously allowed for gke testing during e2e, but
we are going to use kind.  This removes a couple of YAML
stanzas in the test data that kind does not produce.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,7 +43,7 @@ test/pkg:
 test/verify:
 	bazel test //hack/...
 
-# This target uses kind to start a cluster and runs the e2e tests
+# This target uses kind to start a k8s cluster  and runs the e2e tests
 # against that cluster.
 .PHONY: test/e2e
 test/e2e: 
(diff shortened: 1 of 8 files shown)


chrislovecnm:
@chrisseto PTAL

Makefile
```diff
@@ -43,7 +43,7 @@ test/pkg:
 test/verify:
 	bazel test //hack/...
 
-# This target uses kind to start a cluster and runs the e2e tests
+# This target uses kind to start a k8s cluster  and runs the e2e tests
```
chrisseto, line 46: nit: extra space.

e2e/testdata/TestCreatesInsecureCluster/creates_3-node_insecure_cluster.golden
```diff
@@ -15,7 +15,6 @@ spec:
 apiVersion: v1
 kind: Pod
 metadata:
-  creationTimestamp: null
```
chrisseto, line 18: I don't see anything that would obviously result in these lines being removed?

chrisseto (approved):
LGTM! Just a question about the golden changes?

e2e/testdata/TestCreatesInsecureCluster/creates_3-node_insecure_cluster.golden
chrislovecnm: It is how the test is failing.  It is a difference between GKE and Kind :(

PR #151, opened 2020-08-24
Add update code

chrisseto:
Looks like some extra go.mod changes snuck in? Uncertain if that's a problem

### DIFF
--- a/hack/build/repos.bzl
+++ b/hack/build/repos.bzl
@@ -4185,3 +4185,19 @@ def go_repositories():
         sum = "h1:aykwPMVyQyncZ8iLNVMXgJ1l3c6W0+LSOPmqp8JdCjs=",
         version = "v3.1.1",
     )
+    go_repository(
+        name = "com_github_cenkalti_backoff",
+        build_file_generation = "on",
+        build_file_proto_mode = "disable",
+        importpath = "github.com/cenkalti/backoff",
+        sum = "h1:tNowT99t7UNflLxfYYSlKYsBpXdEet03Pg2g16Swow4=",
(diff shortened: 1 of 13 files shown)


chrislovecnm:
CI failed

hack/build/repos.bzl
```diff
@@ -4185,3 +4185,19 @@ def go_repositories():
...
     )
+    go_repository(
+        name = "com_github_cenkalti_backoff",
+        build_file_generation = "on",
+        build_file_proto_mode = "disable",
+        importpath = "github.com/cenkalti/backoff",
+        sum = "h1:tNowT99t7UNflLxfYYSlKYsBpXdEet03Pg2g16Swow4=",
+        version = "v2.2.1+incompatible",
+    )
+    go_repository(
```
chrislovecnm, line 4196: Need to remove this since we do not need this.

hack/update-deps.sh
```diff
@@ -47,13 +47,13 @@ update_bazel=(
...
-  "$jq"
-)
+#update_deps_licenses=(
+#  $(realpath "$6")
+#  "$go"
+#  "$jq"
+#)
 
-shift 6
+shift 5
```
chrislovecnm, line 56: Need to double check this.

chrislovecnm:
> Looks like some extra go.mod changes snuck in? Uncertain if that's a problem

This is fixed.  We had the wrong version of semver imported.  And I included some code from the k8s codebase.  Having 20 lines of new code is better than importing all of k8s.  Probably should move it into its own code file though.

PR #154, opened 2020-08-31
RedHat testing image name override

chrislovecnm:
The statefulset builder now checks if the environment variable
RELATED_IMAGE_COCKROACH is set.  If RELATED_IMAGE_COCKROACH is
set the container image is set to that value instead of the
value set in the CR. RedHat testing sets this ENV variable
to test a specific image from a specific container registry.


keith-mcclellan:
LGTM

PR #173, opened 2020-09-14
Using HOSTNAME ENV variable

chrislovecnm:
Instead of using the hostname -f command to determine the Pod's
hostname. We are using the HOSTNAME environment variable in the
cockroach command.  Various K8s distros and container distros
cannot use hostname -f.

### DIFF
--- a/pkg/resource/statefulset.go
+++ b/pkg/resource/statefulset.go
@@ -198,6 +198,14 @@ func (b StatefulSetBuilder) MakeContainers() []corev1.Container {
 					Name:  "COCKROACH_CHANNEL",
 					Value: "kubernetes-operator",
 				},
+				{
+					Name: "COCKROACH_POD_NAME",
+					ValueFrom: &corev1.EnvVarSource{
+						FieldRef: &corev1.ObjectFieldSelector{
+							FieldPath: "metadata.name",
+						},
+					},
+				},
 			},
 			Ports: []corev1.ContainerPort{
 				{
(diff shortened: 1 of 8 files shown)


pkg/resource/statefulset.go
```diff
@@ -276,7 +276,7 @@ func (b StatefulSetBuilder) dbArgs() []string {
 		"-ecx",
 		">- exec /cockroach/cockroach start" +
 			" --join=" + b.joinStr() +
-			" --advertise-host=$(hostname -f)" +
+			" --advertise-host=$(echo $HOSTNAME)" +
```
chrisseto, line 279: This may not currently be a concern for the operator, so feel free to push back (but maybe add a note somewhere?)

`hostname -f` is the FQDN while `$HOSTNAME` is just the pod name. This will present issues when looking at multi-region clusters or networking together an operator cluster and a non-operator cluster.

```bash
root@cockroachdb-0:/cockroach# echo $HOSTNAME
cockroachdb-0
root@cockroachdb-0:/cockroach# hostname -f
cockroachdb-0.cockroachdb.us-west-2.svc.cluster.local
```

chrislovecnm: hostname -f does not work on openshift, and `$HOSTNAME` may not work on kind.  E2E with CI failed.  I am currently debugging.

chrisseto:
Looks like the goldens need to be updated?

chrislovecnm:
Closing this as this code moved into the security PR

PR #174, opened 2020-09-17
Implementing Security Context for Pods

chrislovecnm:
The container previously ran as root, and this PR implements security content settings to allow the Pod to run as a non-root user.


pkg/resource/statefulset.go
```diff
@@ -44,6 +44,8 @@ const (
 	RHEnvVar        = "RELATED_IMAGE_COCKROACH"
 )
 
+var userId = ptr.Int64(1042)
```
chrislovecnm, line 47: This number needs to be higher for openshift

pkg/resource/statefulset.go
```diff
@@ -157,6 +160,16 @@ func (b StatefulSetBuilder) makePodTemplate() corev1.PodTemplateSpec {
 			TerminationGracePeriodSeconds: ptr.Int64(60),
 			Containers:                    b.MakeContainers(),
 			AutomountServiceAccountToken:  ptr.Bool(false),
+
+			// Setup the pod to not run as root
+			SecurityContext: &corev1.PodSecurityContext{
+				FSGroup:      ptr.Int64(1142),
```
chrislovecnm, line 166: `FSGroup` may not run out of the box with openshift.  We need to make this configurable or remove.  It requires more testing as well.

chrislovecnm: I moved this into a featuregate, so we can test it later

pkg/resource/statefulset.go
```diff
@@ -91,7 +96,7 @@ func (b StatefulSetBuilder) Build(obj runtime.Object) error {
 			Name: certsDirName,
 			VolumeSource: corev1.VolumeSource{
 				Projected: &corev1.ProjectedVolumeSource{
-					DefaultMode: ptr.Int32(0400),
+					DefaultMode: ptr.Int32(0600),
```
chrislovecnm, line 99: This needs to be behind the feature gate as well

pkg/resource/statefulset.go
```diff
@@ -245,14 +263,32 @@ func (b StatefulSetBuilder) MakeContainers() []corev1.Container {
...
+			})
+	}
+
+	return container
 }
 
 func (b StatefulSetBuilder) secureMode() string {
 	if b.Spec().TLSEnabled {
-		return " --certs-dir=/cockroach/cockroach-certs/"
+		return "--certs-dir=/cockroach/cockroach-certs/"
```
chrisseto, line 289: This is the biggest bit that's jumping out at me 🤔  as I don't see any accompanying change?
Or was there just an additional space there in the first place?

chrislovecnm: hrmmmm

chrislovecnm: I do not think that is called anywhere.  I am trying to get this working in debug mode.

chrislovecnm: It is not called anywhere according to grep.

PR #186, opened 2020-10-01
Partitioned Update

chrislovecnm:
Integration new update code for the operator.
  This is using the code from cockroach cloud to
  do correctly partition update for the database.


e2e/assert.go
```diff
@@ -154,3 +157,59 @@ func testPodsWithPredicate(pods []corev1.Pod, pred func(*corev1.Pod) bool) bool
 func statefulSetIsReady(ss *appsv1.StatefulSet) bool {
 	return ss.Status.ReadyReplicas == ss.Status.Replicas
 }
+
+func requireDatabaseToFunction(t *testing.T, sb testenv.DiffingSandbox, b testutil.ClusterBuilder) {
+	sb.Mgr.GetConfig()
+	podName := fmt.Sprintf("%s-0.%s", b.Cluster().Name(), b.Cluster().Name())
```
chrislovecnm, line 163: This is failing because of SSL certificate issues.  In order for the port forwarder code to work it has to look up a pod name, but the pod name such as crdb-0 is not a host in the SSL certificates.

e2e/e2e_test.go
```diff
@@ -325,3 +326,75 @@ func TestUpgradesMajorVersion19_1To19_2(t *testing.T) {
...
+
+func TestDatabaseFunctionality(t *testing.T) {
+	if testing.Short() {
+		t.Skip("skipping test in short mode.")
+	}
+	testLog := zapr.NewLogger(zaptest.NewLogger(t))
+	actor.Log = testLog
+	sb := testenv.NewDiffingSandbox(t, env)
+	sb.StartManager(t, controller.InitClusterReconcilerWithLogger(testLog))
+	builder := testutil.NewBuilder("crdb").WithNodeCount(1).WithTLS().
```
chrislovecnm, line 386: Need to make this a 3 node cluster

pkg/resource/tls_secret.go
```diff
@@ -94,6 +94,14 @@ func (s *TLSSecret) UpdateCertAndCA(cert, ca []byte, log logr.Logger) error {
 	return err
 }
 
+func (s *TLSSecret) CA() []byte {
```
chrislovecnm, line 97: TODO - check which one of these funds are used.  I think one of them is not used.

pkg/update/update_cockroach_version.go
```diff
@@ -348,13 +321,6 @@ func grantRoleToUser(ctx context.Context, db *sql.DB, role roleMembership, usern
 }
 
 func getClusterSetting(ctx context.Context, db *sql.DB, name string) (string, error) {
-	// TODO do we need this?
-	/*
-		if err := isValidClusterSettingName(name); err != nil {
```
chrislovecnm, line 353: Do we need this?? @chrisseto

chrislovecnm:
@chrisseto I to improve the e2e test more, but it is a long PR so PTAL

pkg/actor/upgrade.go
```diff
@@ -129,6 +129,19 @@ func (up *upgrade) Act(ctx context.Context, cluster *resource.Cluster) error {
 	return nil
 }
 
+// TODO this does not handle sha's
+// TODO we are running into the same problem where the version of
+// CR is parsed from the image tag name, we need to fix this
+
+func getImageNameNoVersion(image string) string {
```
chrislovecnm, line 136: This needs to be moved to partitioned update I think.

pkg/features/operator_features.go
```diff
@@ -23,23 +23,26 @@ import (
...
-// beta: v1.0
-//
+	// Every feature gate should add method here following this template:
+	//
+	// // owner: @username
+	// // alpha: v1.4
+	// MyFeature() bool
+
+	// owner: @chrislovecnm
+	// alpha: v0.1
```
chrislovecnm, line 33: these numbers are incorrect, and we need an issue to move this to GA

pkg/actor/actor.go
```diff
@@ -50,13 +52,34 @@ type Actor interface {
...
+	} else {
+		update = newUpgrade(scheme, cl, config)
+	}
+
+	// The order of these actors MATTERS.
+	// We need to have update before deploy so that
+	// updates run before the deploy actor, or
+	// deploy will update the STS container and not
+	// deploy.
+	actor := []Actor{
```
chrisseto, line 75: nit: `actors`

pkg/actor/partitioned_update.go
```diff
@@ -0,0 +1,216 @@
...
+// Act runs a new partitionUpdate.
+// This update pattern handles the sql calls and workflow in order to
+// update a cr cluster.  This is replacing the old update actor.
+func (up *partitionedUpdate) Act(ctx context.Context, cluster *resource.Cluster) error {
+
+	// TODO we have edge cases that we are not covering
+	// see https://github.com/cockroachdb/cockroach-operator/issues/202
+
+	log := up.log.WithValues("CrdbCluster", cluster.ObjectKey())
+	log.Info("checking update opportunities, with using a partitioned Updated")
```
chrisseto, line 69: nit: Either `using` or `with` not both?

pkg/actor/partitioned_update.go
```diff
@@ -0,0 +1,216 @@
...
+	}
+
+	// TODO set status that we are completed.
+	log.Info("update completed with partitioned update", "new version", specStr)
+	CancelLoop(ctx)
+	return nil
+}
+
+// inK8s checks to see if the a file exists
+func inK8s(file string) bool {
```
chrisseto, line 210: Should file just be a constant that's used by this method?

pkg/kube/dialer.go
```diff
@@ -0,0 +1,217 @@
...
+
+func (k *PodDialer) DialContext(ctx context.Context, network, addr string) (net.Conn, error) {
+	return k.Dial(network, addr)
+}
+
+// Dial connects to a port in a kubernetes pod specified by addr. network must be TCP
+// Implmentation adapted from:
+//	https://github.com/kubernetes/kubernetes/blob/27c70773add99e43464a4e525e3bddfc8b602a3d/staging/src/k8s.io/client-go/tools/portforward/portforward.go
+//	https://github.com/kubernetes/kubernetes/blob/27c70773add99e43464a4e525e3bddfc8b602a3d/staging/src/k8s.io/kubectl/pkg/cmd/portforward/portforward.go
+func (k *PodDialer) Dial(network, addr string) (net.Conn, error) {
```
chrisseto, line 129: There are a couple changes to the PodDialer, we found that it was leaking goroutines as `streamConn` is never closed. I'll send you the patch!

pkg/update/update.go
```diff
@@ -283,3 +292,17 @@ func makeWaitUntilAllPodsReadyFuncInAllClusters(
 		return backoff.Retry(f, b)
 	}
 }
+
+// TODO there are ALOT more reason codes in k8sErrors, should we test them all?
```
chrisseto, line 296: Do the k8s errors not provide enough context/information on their own?

pkg/update/update_cockroach_version.go
chrisseto: Nope, that's a sanity check from our code base in case the function ever accepts users input.

I think it's safe to say that it won't ever handle user input in this case.

pkg/update/update_cockroach_version_common.go
```diff
@@ -46,39 +48,66 @@ func makeUpdateCockroachVersionFunction(
...
-				l.Sugar().Infof("%s is running new version on %s", podName, stsNamespace)
+
+				l.Info("is running new version on", "podName", podName, "stsName", stsNamespace)
 				return nil
 			}
 		}
-		return fmt.Errorf("cockroachdb container not found within the cockroach pod")
+
+		// TODO how do we even get here??
+		// I am not certain that this code is even used.
```
chrisseto, line 106: Looks like a sanity check?

Might be more reasonable/understandable to do something like:

```
var container *TheType
for i, c := range pod.Spec.Containers {
    if c.Name == "..." {
        container = &pod.Spec.Containers[i]
    }
}

if container == nil {
  return fmt.Errorf("how did this happen?")
}

// rest of code ...
```

pkg/update/update_cockroach_version_common.go
```diff
@@ -46,39 +48,66 @@ func makeUpdateCockroachVersionFunction(
...
 		stsName := sts.Name
 		stsNamespace := sts.Namespace
 		podName := fmt.Sprintf("%s-%d", stsName, podNumber)
 		clientset := update.clientset
-		crdbPod, err := clientset.CoreV1().Pods(stsNamespace).Get(context.TODO(), stsName, metav1.GetOptions{})
-		if err != nil {
-			return errors.Wrapf(err, "error getting cockroach pod")
+
+		crdbPod, err := clientset.CoreV1().Pods(stsNamespace).Get(update.ctx, podName, metav1.GetOptions{})
+		if k8sErrors.IsNotFound(err) { // this is not an error
```
chrisseto, line 70: This looks like the same code as `handleStsError`?

chrisseto: Why is not found not an error but handled as an error?

pkg/update/update_cockroach_version_common.go
```diff
@@ -46,39 +48,66 @@ func makeUpdateCockroachVersionFunction(
 	}
 }
 
+// TODO some of these should probably be panics or cancel the update at least
```
chrisseto, line 51: I think vlad was using the context to cancel operations previously? A status seems much more explicit IMO.

pkg/update/update_cockroach_version.go
```diff
@@ -152,32 +144,32 @@ func updateClusterStatefulSets(
...
 			numCRDBPods := int(sts.Status.Replicas)
 			if got != numCRDBPods {
-				l.Sugar().Warnf("number of ready replicas is %v, not equal to num CRDB pods %v", got, numCRDBPods)
-				return fmt.Errorf("number of ready replicas is %v, not equal to num CRDB pods %v", got, numCRDBPods)
+				l.Error(err, fmt.Sprintf("number of ready replicas is %v, not equal to num CRDB pods %v", got, numCRDBPods))
+				return err
 			}
 
-			l.Info("all replicas are ready")
+			l.Info("all replicas are ready makeWaitUntilAllPodsReadyFunc update_cockroach_version.go")
```
chrisseto, line 172: Is there not an option to include the lineno and file log messages come from?

pkg/update/update.go
```diff
@@ -74,15 +75,16 @@ type UpdateSts struct {
...
 // a StatefulSet.
 type UpdateTimer struct {
-	podUpdateTimeout          time.Duration
-	podMaxPollingInterval     time.Duration
-	sleeper                   Sleeper
-	waitUntilAllPodsReadyFunc func(context.Context, *zap.Logger) error
+	podUpdateTimeout      time.Duration
+	podMaxPollingInterval time.Duration
+	sleeper               Sleeper
+	// TODO check that this func is actually correct
```
chrisseto, line 81: It might be more correct for it to just check `.Spec.Replicas` against `.Status.ReadyReplicas`?

pkg/database/connection.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+
+	pgCfg.TLSConfig = c.TLSConfig
+	pgCfg.ConnectTimeout = c.ConnectTimeout
+
+	db := stdlib.OpenDB(*pgCfg)
+
+	// Test the database connection
+	if _, err := db.Exec("SELECT 1"); err != nil {
+		return nil, err
```
chrisseto, line 239: nit: I imagine that the errors from this will be incredible opaque. Might be worth tossing in an `errors.Wrap(err, "testing db connection")`

chrisseto (dismissed):
LGTM, couple nits and questions but nothing strictly blocking.

Would be good to fix the PodDialer issues before merging however.

pkg/update/update.go
chrislovecnm: sorta.  The way the code is written we need to act differently depending on the error rather than just assuming that the pod is not found.  For instance, if we get an authz error on an API call, we should fail, and not assume that the pod is not found.  There are ALOT of different types of errors, and the logic needs improving as well in our code.

pkg/update/update_cockroach_version_common.go
chrislovecnm: he was timing out the context after I think 5 minutes, which does not work for updates.

pkg/update/update_cockroach_version.go
chrislovecnm: I have no idea

pkg/update/update.go
chrislovecnm: Legacy code 🤷‍♂️

pkg/actor/actor.go
chrislovecnm: actually, we did not need the variable.  I removed it.

pkg/actor/partitioned_update.go
chrislovecnm: So I wanted to test it, so I moved it to a param, and also it is only used in a single line of code, so I hardcoded it.  Do we want to make it a const?

pkg/kube/dialer.go
chrislovecnm: Please let me know

pkg/update/update_cockroach_version_common.go
chrislovecnm: We talked on slack about this

pkg/database/connection.go
chrislovecnm: Done

pkg/actor/partitioned_update.go
chrislovecnm: Fixed.  Bad English

chrisseto (approved):

Issue #187, opened 2020-10-06
Error with CR v20.1.6


chrislovecnm:
```
$ bazel-bin/hack/bin/kubectl -n crdb-test-mdprkx logs crdb-0
I201006 22:59:31.825740 1 cli/start.go:1223  logging to directory /cockroach/cockroach-data/logs
I201006 22:59:31.826164 1 cli/start.go:1286  CockroachDB CCL v20.1.6 (x86_64-unknown-linux-gnu, built 2020/09/24 18:16:45, go1.13.9)
I201006 22:59:31.826625 1 server/config.go:404  system total memory: 25 GiB
I201006 22:59:31.826665 1 server/config.go:406  server configuration:
max offset             500000000
cache size             6.2 GiB
SQL memory pool size   6.2 GiB
scan interval          10m0s
scan min idle time     10ms
scan max idle time     1s
event log enabled      true
I201006 22:59:31.826675 1 cli/start.go:1122  using local environment variables: COCKROACH_CHANNEL=kubernetes-operator
I201006 22:59:31.826681 1 cli/start.go:1129  process identity: uid 0 euid 0 gid 0 egid 0
I201006 22:59:31.826725 1 cli/start.go:697  starting cockroach node
W201006 22:59:31.826933 47 security/certificate_loader.go:327  could not parse certificate for /cockroach/cockroach-certs/node.crt: empty certificate file: node.crt
E201006 22:59:31.830205 1 cli/error.go:373  ERROR: cannot load certificates.
Check your certificate settings, set --certs-dir, or use --insecure for insecure clusters.

failed to start server: problem using security settings: empty certificate file: node.crt
ERROR: cannot load certificates.
Check your certificate settings, set --certs-dir, or use --insecure for insecure clusters.

failed to start server: problem using security settings: empty certificate file: node.crt
Failed running "start"
```

This seems to happen other 20.1 versions as well.

chrislovecnm:
This is a weird problem that I am having in my dev branch.

PR #188, opened 2020-10-06
New e2e test for CR v20.1.6

chrislovecnm:
This should create the problem that I see in https://github.com/cockroachdb/cockroach-operator/issues/187

I am not certain if this is a kind issue or a problem with the operator.  I am guessing that it is an issue with kind and or our e2e testing.

### DIFF
--- a/e2e/e2e_test.go
+++ b/e2e/e2e_test.go
@@ -168,6 +168,35 @@ func TestCreatesSecureClusterWithGeneratedCert(t *testing.T) {
 	steps.Run(t)
 }
 
+func TestCreatesSecureClusterWithGeneratedCertCRv20(t *testing.T) {
+	if testing.Short() {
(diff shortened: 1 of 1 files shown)


chrislovecnm:
This seems to be a problem with starting a 3 node cluster during e2e tests.

chrislovecnm:
This is a good e2e should we should merge this, but I am having a problem in my branch.

chrisseto (approved):
Test LGTM, I'm wondering if [this method](https://github.com/cockroachdb/cockroach-operator/blob/1759713726c26ee09a1599ded7464d31c181b444/pkg/resource/tls_secret.go#L57-L70) needs to be updated to check for the emptiness of the keys as well as the existence of them?

I'm not seeing anything that would result in an empty key however 🤔

PR #194, opened 2020-10-13
fix e2e tests for macos

alinadonisa:
1. Added  dev/syncdeps target on makefile
2. Rebuild and replace binaries for kubetest2 and kubetest2_kind
3. We still need core-utils for macos install because of the usage of realpath

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,9 @@
 #
 
 # values used in workspace-status.sh
+
 DOCKER_REGISTRY?=us.gcr.io/chris-love-operator-playground
+# DOCKER_REGISTRY?=gcr.io/og-lab-project/alina-playground
 DOCKER_IMAGE_REPOSITORY?=cockroach-operator
 APP_VERSION?=v1.0.0-alpha.1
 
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -96,3 +104,15 @@ release/image:
...
 		//:push_operator_image 
+
+#
+# Dev target that
+#
+.PHONY: dev/syncdeps
+dev/syncdeps:
+	bazel run //hack:update-deps \
+	bazel run //hack:update-bazel \
+	bazel run //:gazelle -- update-repos -from_file=go.mod
```
chrislovecnm, line 115: What causes use to need the `update-repos`?  Also, should we have just an update-bazel target?  I have to do that all the time

chrislovecnm (commented):
Just some nit picks

Makefile
alinadonisa: I had a problem with loading the external dependencies when I first run the make targets. On fresh install on my machine until I did not add -- update-repos, it did not worked.

chrisseto (approved):
LGTM!

Note to myself for future reference `select` can be used for selecting different binaries.

PR #220, opened 2020-11-06
fix certs path in readme

zmalik:
small nit in readme when running any command in built-in SQL client 

```
#
# Welcome to the CockroachDB SQL shell.
# All statements must be terminated by a semicolon.
# To exit, type: \q.
#
ERROR: cannot load certificates.
Check your certificate settings, set --certs-dir, or use --insecure for insecure clusters.
```

but works with `/cockroach/cockroach-certs`


chrislovecnm (approved):
LGTM

PR #231, opened 2020-11-11
point operator.yaml to dockerhub by default

keith-mcclellan:
quick manifest change to point at cockroach-operator in dockerhub


piyush-singh (approved):
LGTM. Note that we are publishing the master branch here. We have clearly labeled this as a beta in our docs. We will need to revisit this with John K, Chris L, and team tomorrow.

PR #233, opened 2020-11-11
Fix image generation for OpenShift

alinadonisa:
- I added a patch  for env var RELATED_IMAGE_COCKROACH that will be used on OpenShift generation.
- I removed automation draft file for OpenShift
- cleanup .gitignore


chrislovecnm (dismissed):
LGTM

chrisseto (approved):

taroface:
For `example.yaml`, I believe we should keep the default resource values I merged yesterday in https://github.com/cockroachdb/cockroach-operator/pull/237/files.

chrisseto (approved):

Issue #234, opened 2020-11-11
K8s Operator guide default resource limits in example.yaml do not work for recommended machine types


piyush-singh:
In our [K8s operator](https://www.cockroachlabs.com/docs/stable/orchestrate-cockroachdb-with-kubernetes.html#install-the-operator) docs for GKE, we have a section that covers resource limits. Our guide walks users through creating a GKE cluster with the machine type as `n1-standard-4`. 

`gcloud container clusters create cockroachdb --machine-type n1-standard-4`

In the resource limits section, we guide the user to set resource limits on their cluster, however we do not recommend values that will work with the `n1-standard-4` machine type.

```
resources:
    requests:
      cpu: "4"
      memory: "16Gi"
    limits:
      cpu: "4"
      memory: "16Gi"
```

This results in errors when trying to apply the `example.yaml` file to the cluster:

```
Events:
 Type   Reason      Age         From        Message
 ----   ------      ----        ----        -------
 Warning FailedScheduling 31s (x12 over 13m) default-scheduler 0/3 nodes are available: 3 Insufficient cpu, 3 Insufficient memory.
```

We should instead recommend the user updates their resource limits to `cpu: "2"`  and `memory: "8Gi"` so that their example file will work with the suggested machine type in the guide. 

Note: Google lists the hardware resources for these machines as:
![image](https://user-images.githubusercontent.com/21692034/98762994-fbbddf00-23a6-11eb-8117-d4d784d11a9d.png)

So there are only 15 GB of ram available. Google does list 4 vCPUs, however, so I'm unclear why K8s is unable to schedule due to the CPU limit. Dropping to 2 CPUs was required in order to get my cluster to spin up successfully though.

johnrk:
@chrislovecnm ,

As discussed in Slack, @DuskEagle and @chrisseto have recommended that users continue using **n1-standard-4**. This would be consistent with CC SRE best practices. And for the example.yaml to be updated with the following:

```
resources:
    requests:
      cpu: "4"
      memory: "15Gi"
```

johnrk:
@piyush-singh , to clarify, this is an Engineering issue, so I've assigned it to Chris Love.

piyush-singh:
[...]

PR #237, opened 2020-11-11
update default resource values

taroface:
Fixes https://github.com/cockroachdb/cockroach-operator/issues/234. These settings were verified to work with `n1-standard-4`.


johnrk (approved):
LGTM

Issue #91, opened 2020-07-17
Operator isn't listening for changes in the dataStore custom resource (e.g.  storage: 50Gi->60Gi)


johnrk:
After successfully creating a 3 node cluster with **config/examples/3-nodes-secure-kube-cert.yaml** , I changed modified the Storage field under dataStore from 60Gi to 50Gi. I expected the persistent volumes and persistent volume claims to change to my desired state, but they didn't.

Screenshots:
<img width="432" alt="Screen Shot 2020-07-17 at 1 25 15 PM" src="https://user-images.githubusercontent.com/1663752/87814338-8e9fa000-c831-11ea-9f74-46caea97eaef.png">
<img width="1257" alt="Screen Shot 2020-07-17 at 1 25 38 PM" src="https://user-images.githubusercontent.com/1663752/87814335-8ba4af80-c831-11ea-8e02-150be8f0dafa.png">

chrislovecnm:
Need to research this, but this may not be feasible on all platforms. This is not a big but a feature that is not implemented

chrislovecnm:
Does the helm chart support this?

chrislovecnm:
This is not supported under all use cases. A list of supported platforms:

> AWS-EBS, GCE-PD, Azure Disk, Azure File, Glusterfs, Cinder, Portworx, and Ceph RBD

Not certain this list is the full list, but we need to mention this.

We also need to determine how this impact database performance?

chrisseto:
@johnrk increasing disk sizes should be supported but decreasing them isn't something the operator (Or kubernetes for that matter) can do.

johnrk:
@chrislovecnm , can you clarify the steps a user is supposed to take to increase disk size?

Right now, it's successfully installing a CRDB cluster with the correct disk size. But it still doesn't appear to change when I do kubectl edit Crdbcluster <cluster name> or editing the yaml file.

In addition, I'm trying to delete the pv,pvc --all when the pods are still running, but it appears to hang.

chrislovecnm:
@johnrk I have not tested it TBH.  All you theoretically should have to do is edit the CRD.

chrisseto:
The operator may have to restart all the CRDB pods? Once a PVC is resized, the actual FS won't be resized until the pod it's attached to is restarted/recreated.

chrislovecnm:
cc @alinadonisa

johnrk:
@taroface , this issue is still open.

PR #244, opened 2020-11-16
Implementing resizing PVCs

chrislovecnm:
This PR includes another actor that resizes PVCs when a CR
is updated.

Fixes https://github.com/cockroachdb/cockroach-operator/issues/91


pkg/actor/resize_pvc.go
```diff
@@ -0,0 +1,144 @@
...
+		if pvcsToKeep[pvc.Name] {
+			pvc.Spec.Resources.Requests.Storage().Set(cluster.Spec().DataStore.VolumeClaim.PersistentVolumeClaimSpec.Resources.Requests.Storage().MilliValue())
+			if _, err := clientset.CoreV1().PersistentVolumeClaims(cluster.Namespace()).Update(ctx, &pvc, metav1.UpdateOptions{}); err != nil {
+				return errors.Wrap(err, "error resizing PVCs")
+			}
+		}
+	}
+
+	log := rp.log.WithValues("CrdbCluster", cluster.ObjectKey())
+	log.Info("resized all PVCs")
```
chrisseto, line 141: Don't we also need to restart all the CRDB pods to get the PVCs to resize as well as modifying the statefulset's template to include the increased size?

chrislovecnm: Yes we do.  Forgot to make this WIP.

pkg/actor/resize_pvc.go
```diff
@@ -0,0 +1,222 @@
...
+	sleeper := update.NewSleeper(1 * time.Minute)
+
+	k8sCluster := &update.UpdateCluster{
+		Clientset:             clientset,
+		PodUpdateTimeout:      podUpdateTimeout,
+		PodMaxPollingInterval: podMaxPollingInterval,
+		Sleeper:               sleeper,
+	}
+
+	return update.RollingRestart(ctx, updateRoach, k8sCluster, rp.log)
```
zmalik, line 221: hey @chrislovecnm this feature is great!
but from v1.15, kubernetes feature `ExpandInUsePersistentVolumes` is in beta which doesn't require pods to be restarted. This avoids unnecessary disruption when expanding the disks. And is supported by GKE already for example. Can we make the restart optional and go with the option of only updating the sts definition? which we can achieve using a delete `--cascade=false` and recreating.

This would provide the necessary resilience that we expect from our cockroachDB clusters

chrisseto: Thanks for the suggestion @zmalik! We can certainly look into adding support for `ExpandInUsePersistentVolumes`.

> This would provide the necessary resilience that we expect from our cockroachDB clusters

A rolling update should be safe as long as you have > 3 nodes in your CockroachDB cluster. Have you experienced otherwise? If so, I'd love to hear any additional details you might have to share!

chrislovecnm: I do not think we need to make it optional, I would recommend using the beta feature, and only supporting 1.15 and up. @chrisseto thoughts?

chrisseto: +1, it seems easier to implement and easier to use.

zmalik: awesome!

> A rolling update should be safe as long as you have > 3 nodes in your CockroachDB cluster

you are right but it still needs the connections to be rebalanced and I know we should be ready for any unexpected downtime of a node but if we can avoid it, its even better. And only use a restart and bounce the connections when there is actually no other way around.

chrislovecnm: @zmalik so I tested this by hand and noticed a couple of things.  We cannot resize the statefulset directly. I am still getting the status that the pods have to be restarted, but that goes away.  Kinda weird.  The volume does resize.

Now, this is only true for some storage classes, so we do need the code for storage classes, that won't resize, and an API value to enable resizing without Pod restarts as well.

@chrisseto do you want another API value, or should the user restart the Pods?

1. Have API value that allows pods to skip restart on resize
2. Add code to not have statefulset volume template update
3. Modify checks in code that only tests the PVC size vs the cluster CR, because we cannot trust the statefulset
4. Figure out if need to modify the volumeclaimtemplate.

```bash
kubectl edit pvc <name> # for each PVC in the StatefulSet, to increase its capacity.
kubectl delete sts --cascade=false # <name> to delete the StatefulSet and leave its pods.
kubectl apply -f <name> # to recreate the StatefulSet.
kubectl rollout restart sts <name> # to restart the pods, one at a time. During restart, the pod's PVC will be resized.
```

zmalik: > so we do need the code for storage classes, that won't resize

@chrislovecnm right, that why I said to make it optional so we can introduce a flag like `restartPodsOnResize` which can be true by default, but users which have a storage-class which supports resizing without restart can set it to `true` 

to modify `volumeclaimtemplate` unfortunately there is no way right now except then recreating the sts https://github.com/kubernetes/kubernetes/issues/68737

we are deleting the sts using `cascade=false` and recreating it, which doesn't restart the pods either.

pkg/actor/decommission.go
```diff
@@ -57,7 +57,13 @@ func (d decommission) Handles(conds []api.ClusterCondition) bool {
 }
 
 func (d decommission) Act(ctx context.Context, cluster *resource.Cluster) error {
+
 	log := d.log.WithValues("CrdbCluster", cluster.ObjectKey())
+
+	if !utilfeature.DefaultMutableFeatureGate.Enabled(features.Decommission) {
```
chrislovecnm, line 63: We missed the feature gate for decommision.  I moved them into the actors so that building the slice is easier.

pkg/actor/resize_pvc.go
chrislovecnm: @zmalik PTAL.

chrislovecnm:
@chrisseto PTAL
@alinadonisa PTAL

pkg/actor/decommission.go
alinadonisa: @chrislovecnm The decommission flag  it is checked in the Handle method. If Handle returns true than Act. No need for this condition on Act.

chrislovecnm:
Need to make the new API value optional.

chrislovecnm:
TODO - need to not true to run PVC resize if the cluster is not ready.

chrislovecnm:
I have a bug in GKE, even though the e2e with kind is passing.  Arg

pkg/actor/resize_pvc.go
```diff
@@ -0,0 +1,246 @@
...
+	}
+	if err := cluster.Spec().DataStore.Apply(resource.DataDirName,
+		resource.DbContainerName, resource.DataDirMountPath, &newSts.Spec, metaMutator); err != nil {
+
+		return err
+	}
+
+	// delete the original sts, but do not delete the Pods
+	orphan := metav1.DeletePropagationOrphan
+	if err := rp.client.Delete(ctx, sts, &client.DeleteOptions{PropagationPolicy: &orphan}); err != nil {
```
zmalik, line 162: 👏

pkg/actor/decommission.go
chrislovecnm: Awesome! I will change mine to match

config/crd/bases/crdb.cockroachlabs.com_crdbclusters.yaml
```diff
@@ -255,6 +255,12 @@ spec:
                           type: string
                       type: object
                   type: object
+                supportsAutoResize:
```
zmalik, line 258: @chrislovecnm the flag name can be confusing.
I mean from code I see it just does a rolling restart of sts if it's true. But actual resize of PVC is supported by default, independent of this flag. I think this can be confusing, can we rename it to something more closer to its implementation like `restartPodsOnResize`?

chrislovecnm: I am not talented with naming, but we need something about that the PVC supports pods on restart. Names are hard 😆 

@chrisseto PRTAL

pkg/actor/resize_pvc.go
```diff
@@ -0,0 +1,248 @@
...
+		return err
+	}
+
+	// Use same StatefulSetBuilder that we run in Deploy to
+	// rebuild and save the StatefulSet with the new PVC size
+	r := resource.NewManagedKubeResource(ctx, rp.client, cluster, kube.AnnotatingPersister)
+	_, err := (resource.Reconciler{
```
chrislovecnm, line 159: I am getting:

```
2020-12-01T00:24:15.372Z	ERROR	controller.CrdbCluster	action failed	{"CrdbCluster": "default/cockroachdb", "error": "updating statefulset default.cockroachdb: failed to reconcile statefulset: StatefulSet.apps \"cockroachdb\" is invalid: spec: Forbidden: updates to statefulset spec for fields other than 'replicas', 'template', and 'updateStrategy' are forbidden", "errorVerbose": "updating statefulset default.cockroachdb: failed to reconcile statefulset: StatefulSet.apps \"cockroachdb\" is invalid: spec: Forbidden: updates to statefulset spec for fields other than 'replicas', 'template', and 'updateStrategy' are forbidden\n(1) attached stack trace\n  -- stack trace:\n  | github.com/cockroachdb/cockroach-operator/pkg/actor.(*resizePVC).Act\n  | \tpkg/actor/resize_pvc.go:126\n  | [...repeated from below...]\nWraps: (2) updating statefulset default.cockroachdb\nWraps: (3) attached stack trace\n  -- stack trace:\n  | github.com/cockroachdb/cockroach-operator/pkg/actor.(*resizePVC).updateSts\n  | \tpkg/actor/resize_pvc.go:170\n  | github.com/cockroachdb/cockroach-operator/pkg/actor.(*resizePVC).Act\n  | \tpkg/actor/resize_pvc.go:125\n  | github.com/cockroachdb/cockroach-operator/pkg/controller.(*ClusterReconciler).Reconcile\n  | \tpkg/controller/cluster_controller.go:101\n  | sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler\n  | \texternal/io_k8s_sigs_controller_runtime/pkg/internal/controller/controller.go:256\n  | sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n  | \texternal/io_k8s_sigs_controller_runtime/pkg/internal/controller/controller.go:232\n  | sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).worker\n  | \texternal/io_k8s_sigs_controller_runtime/pkg/internal/controller/controller.go:211\n  | k8s.io/apimachinery/pkg/util/wait.BackoffUntil.func1\n  | \texternal/io_k8s_apimachinery/pkg/util/wait/wait.go:155\n  | k8s.io/apimachinery/pkg/util/wait.BackoffUntil\n  | \texternal/io_k8s_apimachinery/pkg/util/wait/wait.go:156\n  | k8s.io/apimachinery/pkg/util/wait.JitterUntil\n  | \texternal/io_k8s_apimachinery/pkg/util/wait/wait.go:133\n  | k8s.io/apimachinery/pkg/util/wait.Until\n  | \texternal/io_k8s_apimachinery/pkg/util/wait/wait.go:90\n  | runtime.goexit\n  | \tsrc/runtime/asm_amd64.s:1373\nWraps: (4) failed to reconcile statefulset\nWraps: (5) StatefulSet.apps \"cockroachdb\" is invalid: spec: Forbidden: updates to statefulset spec for fields other than 'replicas', 'template', and 'updateStrategy' are forbidden\nError types: (1) *withstack.withStack (2) *errutil.withPrefix (3) *withstack.withStack (4) *errutil.withPrefix (5) *errors.StatusError"}
```

Here.

config/crd/bases/crdb.cockroachlabs.com_crdbclusters.yaml
chrisseto: `restartPodsOnResize` sounds great!

How about ''Set to true if ExpandInUsePersistentVolumes is enabled on kubelet; volumes will be resized without pod disruption" as the description?

pkg/actor/resize_pvc.go
```diff
@@ -0,0 +1,248 @@
...
+		Name:      cluster.StatefulSetName(),
+	}
+	statefulSet := &appsv1.StatefulSet{}
+	if err := rp.client.Get(ctx, key, statefulSet); err != nil {
+		return errors.Wrap(err, "failed to fetch statefulset")
+	}
+
+	// TODO statefulSetIsUpdating is not quite working as expected.
+	// I had to check status.  We should look at the update code in partition update to address this
+	if statefulSetIsUpdating(statefulSet) {
```
chrisseto, line 95: Would this be better/possible to check in `Handles`?

pkg/actor/resize_pvc.go
chrisseto: I think we talked about this? but you have to poll to ensure the sts is fully deleted before recreating it.

The foreground propagation policy is mutually exclusive with orphan for some reason.

pkg/actor/resize_pvc.go
```diff
@@ -0,0 +1,248 @@
...
+	if err := rp.findAndResizePVC(ctx, statefulSet, cluster, clientset); err != nil {
+		return errors.Wrapf(err, "updating PVCs for statefulset %s.%s", cluster.Namespace(), cluster.StatefulSetName())
+	}
+
+	log.Info("Starting updating sts")
+
+	// Update the STS with the correct volume size, in case more pods are created
+	// We will create a copy and update the copy, and then delete the original without
+	// deleting the Pods.  The new sts is then used to create a new statefulset.
+	if err := rp.updateSts(ctx, statefulSet, cluster); err != nil {
```
chrisseto, line 137: Should we update the sts before resizing the pvcs? Seems like it's possible to hit a race condition that results in some pods not getting resized if the sts replicas change at the same time.

chrislovecnm: The sts should not get updated at the same time and I am torn where an how to do this :(

config/crd/bases/crdb.cockroachlabs.com_crdbclusters.yaml
zmalik: > How about ''Set to true if ExpandInUsePersistentVolumes is enabled on kubelet

but that would mean host access to kubelet right? I'm not aware that kubernetes API exposes the enabled feature gates

pkg/actor/resize_pvc.go
chrislovecnm: Yep, we talked about it.  I have a retry with a backoff implemented now.

pkg/actor/resize_pvc.go
chrislovecnm: Yep.  I am planning on moving it.

pkg/actor/resize_pvc.go
```diff
@@ -0,0 +1,269 @@
...
+	if err := rp.client.Delete(ctx, sts, &client.DeleteOptions{PropagationPolicy: &orphan}); err != nil {
+		return err
+	}
+
+	f := func() error {
+		return rp.recreateSTS(ctx, cluster)
+	}
+
+	b := backoffFactory(5 * time.Minute)
+	return backoff.Retry(f, backoff.WithContext(b, ctx))
```
chrislovecnm, line 173: @chrisseto, here is my backoff retry for creating the new STS.  Once K8s has reconciled the deletion of the sts it creates the updated version.

pkg/update/update.go
```diff
@@ -193,13 +193,17 @@ func PartitionedRollingUpdateStrategy(perPodVerificationFunc func(*UpdateSts, in
...
 
-			// If pod already updated, we are probably retrying a failed job
-			// attempt. Best not to redo the update in that case, especially the sleeps!!
-			if err := perPodVerificationFunc(updateSts, int(partition), l); err == nil {
-				l.Info("already updated, skipping sleep", "partition", partition)
-				skipSleep = true
-				continue
-			}
+			/*
+				// TODO why are we doing this?
```
chrislovecnm, line 197: @chrisseto @alinadonisa I am trying to reuse the update code in order to do a rolling update of the cluster.  So two things are happening.  

1. This code block is stopping the update from happening
2. The sts is not updating so we are not rolling the pods.

So ideas
1. don't reuse the update code
2. update the sts so that it does an update
3. change the code somehow so we are inserting the capability to delete a pod

Thoughts?

pkg/update/rolling_restart.go
```diff
@@ -0,0 +1,108 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// TODO this is not working.
+// We need to add an annotation the statefulset with a value that will
+// cause the pods to roll.
+// See https://github.com/kubernetes/kubernetes/blob/bd4d197b5267ee198c1b0a070d7398f10df68c52/staging/src/k8s.io/kubectl/pkg/polymorphichelpers/objectrestarter.go#L95
+
+// RollingRestart performs a rolling restart on a sts.
+func RollingRestart(
```
chrisseto, line 36: Should this be commented out or removed until it's working?

chrisseto (changes requested):

chrisseto:
Quick PSA if you're testing on EKS: [_Sometimes_ resizing fails due to a bug in EC2's API](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/622).

pkg/actor/resize_pvc.go
chrislovecnm: Looking at the code, I am not certain we can move it from here.  We need to get status working to help with this.

pkg/actor/resize_pvc.go
```diff
@@ -0,0 +1,280 @@
...
+	clientset, err := kubernetes.NewForConfig(rp.config)
+	if err != nil {
+		return errors.Wrapf(err, "failed to create kubernetes clientset")
+	}
+
+	stsStorageSizeDeployed := statefulSet.Spec.VolumeClaimTemplates[0].Spec.Resources.Requests.Storage()
+	stsStorageSizeSet := cluster.Spec().DataStore.VolumeClaim.PersistentVolumeClaimSpec.Resources.Requests.Storage()
+
+	// If the sizes match do not resize
+	if stsStorageSizeDeployed.Equal(stsStorageSizeSet.DeepCopy()) {
```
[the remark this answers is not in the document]
chrislovecnm: You lost me.  I am not certain what you mean.

pkg/actor/resize_pvc.go
```diff
@@ -0,0 +1,280 @@
...
+	log.Info("Starting updating sts")
+
+	// Update the STS with the correct volume size, in case more pods are created
+	// We will create a copy and update the copy, and then delete the original without
+	// deleting the Pods.  The new sts is then used to create a new statefulset.
+	if err := rp.updateSts(ctx, statefulSet, cluster); err != nil {
+		return errors.Wrapf(err, "updating statefulset %s.%s", cluster.Namespace(), cluster.StatefulSetName())
+	}
+
+	// TODO this is not working so we will need to patch the sts
```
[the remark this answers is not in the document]
chrislovecnm: I am wondering a couple of things.  

1. Have RestartClusterOnResize flag, since I think most of the times we do not need to roll the cluster.
2. Yes this code is not working as we need to patch the sts before doing the rolling restart.

pkg/actor/resize_pvc.go
```diff
@@ -0,0 +1,280 @@
...
+	}
+
+	log.Info("resizing PVCs")
+	for _, pvc := range pvcs.Items {
+		// Resize PVCs that are still in use
+		if pvcsToKeep[pvc.Name] {
+			size := cluster.Spec().DataStore.VolumeClaim.PersistentVolumeClaimSpec.Resources.Requests.Storage()
+			pvc.Spec.Resources.Requests[v1.ResourceStorage] = *size
+			//pvc.Spec.Resources.Requests.Storage().Set(cluster.Spec().DataStore.VolumeClaim.PersistentVolumeClaimSpec.Resources.Requests.Storage().Value())
+			log.Info(fmt.Sprintf("size %s", pvc.Spec.Resources.Requests.Storage().String()))
```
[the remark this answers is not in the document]
chrislovecnm: Nope no, Infof.  The logging library is weird.  We need to start doing logging levels.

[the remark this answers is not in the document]
chrislovecnm: Also, I need to remove this line, it was logging for debugging.

pkg/update/rolling_restart.go
chrislovecnm: Commented, but the code works, but it does not roll the cluster because we are not patching the STS.  We patch the sts and this code will work.

pkg/actor/resize_pvc.go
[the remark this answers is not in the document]
chrislovecnm: @chrisseto ping

manifests/operator.yaml
```diff
@@ -222,8 +222,13 @@ spec:
         - name: cockroach-operator
           image: cockroachdb/cockroach-operator
           imagePullPolicy: IfNotPresent
+<<<<<<< HEAD
```
chrisseto, line 225: left over merge conflicts?

pkg/actor/resize_pvc.go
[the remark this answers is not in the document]
chrisseto: In a naive read through of this code, this cases jumps out to me. It looks like Some PVCs might not get resized if something goes wrong.
If the order was 1. Update STS, 2. Update PVCs; this check could result in some PVCs not being updated.

To ease some future confusion, it may be helpful to update this comment point out that updating the STS update is performed last?

chrisseto (dismissed):
LGTM just need to remove the leftover merge conflict marker 👍

manifests/operator.yaml
chrislovecnm: Fixed, but we are getting CI issues.

chrislovecnm:
@chrisseto can you take a look at this again?  I thought this was merged

chrisseto (dismissed):

keith-mcclellan:
LGTM, and @chrisseto gave this the thumbs up.  What do we need to do to get this merged?

chrisseto (approved):

Issue #137, opened 2020-08-12
PDB Issue: minAvailable is incorrect


johnrk:
**Test Case:**

- minAvailable: 2
- maxUnavailable: 1

**Expected Result:** 
when running kubectl get poddisruptionbudget, it would return minAvailable: 2

**Actual Result:**
<img width="602" alt="Screen Shot 2020-08-12 at 2 03 58 PM" src="https://user-images.githubusercontent.com/1663752/90050786-b48b5980-dca4-11ea-8bbe-54f417ed9b4c.png">

**Full yaml:**
<img width="586" alt="Screen Shot 2020-08-12 at 2 05 02 PM" src="https://user-images.githubusercontent.com/1663752/90050878-d684dc00-dca4-11ea-9532-817d3811e6ee.png">

chrislovecnm:
cc @alinadonisa

PR #263, opened 2020-12-16
fix minAvailable typo for  PDB issue

alinadonisa:
- fix https://github.com/cockroachdb/cockroach-operator/issues/137
- regenerate crd and csv for OpenShift with the fix


bundle/manifests/cockroach-operator.clusterserviceversion.yaml
```diff
@@ -304,5 +304,4 @@ spec:
   provider:
     name: Cockroach Labs
     url: https://www.cockroachlabs.com
-  replaces: cockroach-operator.v1.1.2
```
chrisseto, line 307: Is it expected that this value was removed?

chrisseto (approved):
LGTM!

PR #279, opened 2021-01-08
Update statefulset.go

chrisseto:


pkg/resource/statefulset.go
```diff
@@ -182,9 +182,11 @@ func (b StatefulSetBuilder) MakeContainers() []corev1.Container {
...
 	// variable to allow the testing of a specific image.
 	//
-	image := os.Getenv(RHEnvVar)
+	image := b.Spec().Image.Name
 	if image == "" {
-		image = b.Spec().Image.Name
+		if rhe := os.Getenv(RHEnvVar); rhe != "" {
+			image = rhe
+		}
 	}
```
chrislovecnm, line 191:
```suggestion
if image ==  "" {
     return fmt.Error("failed to find image")
}
```

You will need to update the return value.

chrislovecnm (changes requested):
We need to check if both are left blank.

chrisseto:
Closed in favor of https://github.com/cockroachdb/cockroach-operator/pull/293

Issue #206, opened 2020-10-27
Remove emptyDir


chrislovecnm:
CRDB cannot run when using emptyDir.  Check that the PVC that was defined is not an empty dir.

vktr-brlv:
@chrislovecnm Could you please clarify more on the issue? Thanks!

chrislovecnm:
In the api you are able to use an empydir disk type. Remove that and fix the tests to use pvc’s

chrislovecnm:
This is fixed, please test

PR #294, opened 2021-01-21
emtydir removed #206

vktr-brlv:


e2e/e2e_test.go
```diff
@@ -111,7 +111,7 @@ func TestCreatesInsecureCluster(t *testing.T) {
 	sb := testenv.NewDiffingSandbox(t, env)
 	sb.StartManager(t, controller.InitClusterReconcilerWithLogger(testLog))
 
-	b := testutil.NewBuilder("crdb").WithNodeCount(3).WithEmptyDirDataStore()
+	b := testutil.NewBuilder("crdb").WithNodeCount(3).WithHostPathDataStore("/mnt/data", nil)
```
vktr-brlv, line 114: Hey @chrislovecnm ,
I have question here. This test fails b/c kind-test cluster is running single cluster node, and sharing hostpath, so pod failing to launch with error 
`E210120 13:01:42.934643 1 cli/error.go:233  could not cleanup temporary directories from record file: could not lock temporary directory /cockroach/cockroach-data/cockroach-temp150903260, may still be in use: IO error: While lock file: /cockroach/cockroach-data/cockroach-temp150903260/TEMP_DIR.LOCK: Resource temporarily unavailable`
Should we change volume type for this case to pvc? Or maybe run multiple node test cluster? WDYT?

chrislovecnm: Already started to work on this PR

chrislovecnm: https://github.com/cockroachdb/cockroach-operator/pull/266

chrislovecnm: Yes, we have not tested hostDir.  Does it work on a cluster?

vktr-brlv: yes, single node works fine, in multi node container fails with error above when running on single k8s instance with error above. Which makes sense.

chrislovecnm: Can we remove this please?  Let's move this to PVC

chrislovecnm:
@vktr-brlv what is the status on this?

chrislovecnm:
Closing because we did it in #420

PR #299, opened 2021-01-25
replace tls.crt and tls.key to our internal node.crt and node.key nam…

keith-mcclellan:
…ing convention

The database requires the files presented to the database for node to node mTLS be called node.crt and node.key.  For some reason the operator has been generating keys named tls.crt and tls.key and then renaming them as they get projected into the statefulset.  This PR brings us in compliance with the naming standard used by the rest of our tooling.


keith-mcclellan:
@chrislovecnm will this break us requesting a cert from the k8s CA?  Intent here is to just change the default name for the node cert and key in the secret is being injected as so it will match what we would do manually.

vladdy:
@keith-mcclellan Just to give you a bit of background on that, `tls.key` and `tls.crt` are the names used widely for the SSL certificates in Kubernetes (i.e. in `cert-manager` which as it appears becoming a standard). I did not check where the convention is coming from though.

If you change the names, please consider leaving a way to overwrite them as, otherwise, it can become really hard to integrate the operator with Kubernetes native SSL management solutions.

chrislovecnm:
@keith-mcclellan can we close this and do we have an issue open?  We need to access the secret and read the name of the certificate, and then modify the stateful set accordingly, rather than doing what we are doing.

chrislovecnm:
Closing this for now.

PR #301, opened 2021-01-26
fix example.yaml link

taroface:
Fixes https://github.com/cockroachdb/docs/issues/9461.

PR #316, opened 2021-02-06
Fix link to example.yaml

bhearsum:
The current link is a 404.


bhearsum:
I see this is a dupe of #301, whoops!

Issue #305, opened 2021-02-01
Status update improvements


chrislovecnm:
Once we have improved the Status capability via the OpenShift work, all of our actors need to update the status as they do work.  This will allow for better debugging functionality, and also, this will allow different actors to not do operations on a cluster when an error has occurred.  The operator has little functionality that allows the operator to stop working on a cluster, even if an error has occurred.  This can cause issues when say, an upgrade fails.  We do have some error handling, but it is limited.

chrislovecnm:
@keith-mcclellan thoughts?

keith-mcclellan:
how useful would this be outside of an OpenShift based deployment?  If it will work generically it's worth considering, otherwise we should put this in the backlog of future work.

alinadonisa:
@keith-mcclellan , @chrislovecnm  it is already in progress and used on Openshift. For now it is very generic and simple. It sets 3 states: Failed, Starting and Finished. Starting is the state used mainly for Openshift to skip the timeout. Failed it logs the errors from the operator and Finished will signal that the reconciliation of the CRD finished without any error on any steps.

keith-mcclellan:
As long as this work isn't so openshift specific that we end up having to reimplement this logic to work properly in a generic k8s environment I agree this is necessary for our next release

chrislovecnm:
This is needed cross platform

chrislovecnm:
This is complete.  Closing

Issue #286, opened 2021-01-17
Add new cockroachVersion field


alinadonisa:
1. Implement a new API field for CRDB version
2. When we have an image name and not the cockroachVersion API field, we need to create a Job that spins up a container and parses the CRDB version, from cockroach --version.
3. The job is created each time to check the version.

chrislovecnm:
@johnrk this is done. Can we close

PR #319, opened 2021-02-16
Adding CockroachDB Version Validator

alinadonisa:
- Add version checker action: Issue https://github.com/cockroachdb/cockroach-operator/issues/286
- Add cluster Status: Issue https://github.com/cockroachdb/cockroach-operator/issues/305 
- Fix for  //hack:update-codegen


alinadonisa:
It should be into master  not from master.

Issue #304, opened 2021-02-01
Operator logging is not configurable


chrislovecnm:
We need to take all of our logging and utilize logging levels.  At scale, the operator is going to produce way too much logging.  We need to implement and document DEBUG, INFO, WARN, and ERROR logging levels.

https://github.com/cockroachdb/cockroach-operator/blob/bb9aeafd821197863f73224255d12de46ebb12df/pkg/actor/decommission.go#L61

Is a great example where we should have DEBUG logging level and not INFO.  The logging library that we use allows for the implementation of different "v" logging levels.

For instance

https://github.com/go-logr/logr/blob/8fc6c73f3911f11de3d0c18ade9880f1114ae72f/examples/usage_example.go#L66

DEBUG = v 10
INFO = v 8
WARN = v 6
ERROR = v1

The operator should run with WARN as the default logging level.

johnrk:
Here are docs on CockroachDB's existing logging levels: https://www.cockroachlabs.com/docs/stable/debug-and-error-logs.html#severity-levels

johnrk:
This is related to this open bug: https://github.com/cockroachdb/cockroach-operator/issues/175

PR #323, opened 2021-02-22
Logging changes for operator

chrislovecnm:
## Synopsis

This PR implement does two things.  First, we are now able to set the log level on the command line.  The operator now accepts the zap logging flags, and you can set the level to say "debug".  The operator framework uses the zap logging library and the best manner to implement logging, which I found was to use the zap library.  We will set the operator to "warn", but the zap library defaults to debug.

The second change is a general audit of which logging level is used for which logging statement.  Several changes were made to include "debug", "info", "warn", and other logs. 

Fixes: https://github.com/cockroachdb/cockroach-operator/issues/304

## TODO

- [ ] PR review - I need someone who is familiar with how the logs worked in the past.
- [ ] Update operator.yaml for production mode - this file is incorrect currently
- [ ] Backout BUILD.bazel changes - this file needs merging (working around an existing issue we have)
- [ ] Run e2e tests by hand

## Next steps

We need information on how to configure logging via the command line included in the documentation - @johnrk 

## Future improvements

I may have missed a couple of logging statements.  I have done a search, but as we improve the operator we need to ensure that it is logging properly.  We need to determine if we are going to use the zap library in its native implementation.  We are using the logr interface backed by zap, and using zap directly gives you more capability.  This should be backlogged and is a medium-level priority.  - @johnrk

### DIFF
--- a/manifests/BUILD.bazel
+++ b/manifests/BUILD.bazel
@@ -6,8 +6,8 @@ k8s_deploy(
     images = {
         # TODO how do we do this?
         # "{STABLE_DOCKER_REGISTRY}/{STABLE_IMAGE_REPOSITORY}:{STABLE_DOCKER_TAG}"
-        #"cockroachdb/cockroach-operator:v1.16.12-rc.2": "//cmd/cockroach-operator:operator_image",
-        "cockroach-operator:bazel": "//cmd/cockroach-operator:operator_image",
+        # "cockroachdb/cockroach-operator:v1.16.12-rc.2": "//cmd/cockroach-operator:operator_image",
+        "cockroachdb:bazel": "//cmd/cockroach-operator:operator_image",
     },
     template = ":operator.yaml",
 )
--- a/manifests/operator.yaml
+++ b/manifests/operator.yaml
@@ -263,7 +263,7 @@ spec:
           imagePullPolicy: IfNotPresent
           # uncomment this to enable different new features
           # args: ["-feature-gates","PartitionedUpdate=true,UseDecommission=true,Upgrade=false, CrdbVersionValidator=false"]
-          args: ["-feature-gates","Upgrade=false"]
+          args: ["-feature-gates","Upgrade=false","--zap-log-level","info"]
           env:
             - name: WATCH_NAMESPACE
               valueFrom:
--- a/pkg/actor/decommission.go
+++ b/pkg/actor/decommission.go
@@ -30,6 +30,7 @@ import (
 	"github.com/cockroachdb/cockroach-operator/pkg/scale"
 	"github.com/cockroachdb/cockroach-operator/pkg/utilfeature"
 	"github.com/pkg/errors"
+	"go.uber.org/zap/zapcore"
 	appsv1 "k8s.io/api/apps/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	kubetypes "k8s.io/apimachinery/pkg/types"
@@ -64,7 +65,7 @@ func (d decommission) Handles(conds []api.ClusterCondition) bool {
 func (d decommission) Act(ctx context.Context, cluster *resource.Cluster) error {
 
 	log := d.log.WithValues("CrdbCluster", cluster.ObjectKey())
-	log.Info("check decommission oportunities")
+	log.V(int(zapcore.DebugLevel)).Info("check decommission oportunities")
 
 	stsName := cluster.StatefulSetName()
 
@@ -74,13 +75,13 @@ func (d decommission) Act(ctx context.Context, cluster *resource.Cluster) error 
 	}
 	ss := &appsv1.StatefulSet{}
 	if err := d.client.Get(ctx, key, ss); err != nil {
-		log.Info("decommission failed to fetch statefulset")
+		log.Error(err, "decommission failed to fetch statefulset")
 		return kube.IgnoreNotFound(err)
 	}
 	status := &ss.Status
 
 	if status.CurrentReplicas == 0 || status.CurrentReplicas < status.Replicas {
-		log.Info(" decommission statefulset does not have all replicas up")
+		log.V(int(zapcore.WarnLevel)).Info("decommission statefulset does not have all replicas up")
 		return NotReadyErr{Err: errors.New("decommission statefulset does not have all replicas up")}
 	}
 
@@ -88,10 +89,11 @@ func (d decommission) Act(ctx context.Context, cluster *resource.Cluster) error 
 	//We do not scale down if the nodes field is less than 3
 	//TODO @alina add validation webhook (see https://github.com/cockroachdb/cockroach-operator/issues/245)
 	if nodes < 3 {
-		log.Info("We cannot decommission if there are less than 3 nodes", "nodes", nodes)
-		return errors.New("decommission with less than 3 nodes is not supported")
+		err := errors.New("decommission with less than 3 nodes is not supported")
+		log.Error(err, "We cannot decommission if there are less than 3 nodes", "nodes", nodes)
+		return err
 	}
-	log.Info("replicas decommisioning", "status.CurrentReplicas", status.CurrentReplicas, "expected", cluster.Spec().Nodes)
+	log.V(int(zapcore.InfoLevel)).Info("replicas decommisioning", "status.CurrentReplicas", status.CurrentReplicas, "expected", cluster.Spec().Nodes)
 	if status.CurrentReplicas <= cluster.Spec().Nodes {
 		return nil
 	}
@@ -105,10 +107,10 @@ func (d decommission) Act(ctx context.Context, cluster *resource.Cluster) error 
 
 	serviceName := cluster.PublicServiceName()
 	if runningInsideK8s {
-		log.Info("operator is running inside of kubernetes, connecting to service for db connection")
+		log.V(int(zapcore.DebugLevel)).Info("operator is running inside of kubernetes, connecting to service for db connection")
 	} else {
 		serviceName = fmt.Sprintf("%s-0.%s.%s", cluster.Name(), cluster.Name(), cluster.Namespace())
-		log.Info("operator is NOT inside of kubernetes, connnecting to pod ordinal zero for db connection")
+		log.V(int(zapcore.DebugLevel)).Info("operator is NOT inside of kubernetes, connnecting to pod ordinal zero for db connection")
 	}
 
 	// The connection needs to use the discovery service name because of the
@@ -134,7 +136,7 @@ func (d decommission) Act(ctx context.Context, cluster *resource.Cluster) error 
 	if err != nil {
 		return errors.Wrapf(err, "failed to create database connection")
 	}
-	log.Info("opened db connection")
+	log.V(int(zapcore.DebugLevel)).Info("opened db connection")
 	defer db.Close()
 
 	timeout, err := clustersql.RangeMoveDuration(ctx, db)
@@ -168,7 +170,7 @@ func (d decommission) Act(ctx context.Context, cluster *resource.Cluster) error 
 	}
 	// TO DO @alina we will need to save the status foreach action
 	cluster.SetTrue(api.DecommissionCondition)
-	log.Info("decommission completed", "cond", ss.Status.Conditions)
+	log.V(int(zapcore.DebugLevel)).Info("decommission completed", "cond", ss.Status.Conditions)
 	CancelLoop(ctx)
 	return nil
 }
--- a/pkg/actor/deploy.go
+++ b/pkg/actor/deploy.go
@@ -26,6 +26,7 @@ import (
 	"github.com/cockroachdb/cockroach-operator/pkg/resource"
 	"github.com/cockroachdb/cockroach-operator/pkg/utilfeature"
 	"github.com/pkg/errors"
+	"go.uber.org/zap/zapcore"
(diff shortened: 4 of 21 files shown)


manifests/BUILD.bazel
```diff
@@ -6,7 +6,8 @@ k8s_deploy(
     images = {
         # TODO how do we do this?
         # "{STABLE_DOCKER_REGISTRY}/{STABLE_IMAGE_REPOSITORY}:{STABLE_DOCKER_TAG}"
-        "cockroachdb/cockroach-operator:v1.16.12-rc.2": "//cmd/cockroach-operator:operator_image",
+        "cockroachdb:bazel": "//cmd/cockroach-operator:operator_image",
+        # "cockroachdb/cockroach-operator:v1.16.12-rc.2": "//cmd/cockroach-operator:operator_image",
```
chrislovecnm, line 10: This is a bug that needs to be fixed.  Already filed, will back out changes.

manifests/operator.yaml
```diff
@@ -257,11 +257,14 @@ spec:
       serviceAccountName: cockroach-operator-sa
       containers:
         - name: cockroach-operator
-          image: cockroachdb/cockroach-operator:v1.6.12-rc.2
```
chrislovecnm, line 260: This is a bug that needs to be fixed.  Already filed, will back out changes.

manifests/operator.yaml
```diff
@@ -257,11 +257,14 @@ spec:
...
+          image: cockroachdb:bazel
+          # TODO
+          # image: cockroachdb/cockroach-operator:v1.6.12-rc.2
           imagePullPolicy: IfNotPresent
           # uncomment this to enable ParitionedUpdates
-          # args: ["-feature-gates","PartitionedUpdate=true,UseDecommission=true,Upgrade=false"]
-          args: ["-feature-gates","Upgrade=false"]
+          args: ["-feature-gates","PartitionedUpdate=true,UseDecommission=true,Upgrade=false","--zap-log-level","info"]
```
chrislovecnm, line 265: Will back out changes, and also include a new logging level.

keith-mcclellan:
@alyshanjahani-crl could you please work with @chrisseto on reviewing this one?  Thanks!

chrislovecnm:
@keith-mcclellan, I need someone to review this that is familiar with what the operator should output when it is running in production mode.  I would recommend warn or info production logging.  We need to provide just enough information to view what the operator is doing while not flooding the logs with too much information at scale.

pkg/actor/decommission.go
```diff
@@ -68,24 +69,24 @@ func (d decommission) Act(ctx context.Context, cluster *resource.Cluster) error
 	}
 	ss := &appsv1.StatefulSet{}
 	if err := d.client.Get(ctx, key, ss); err != nil {
-		log.Info("decommission failed to fetch statefulset")
+		log.V(int(zapcore.WarnLevel)).Info("decommission failed to fetch statefulset")
```
keith-mcclellan, line 72: This should be Error not Warn

pkg/actor/decommission.go
```diff
@@ -68,24 +69,24 @@ func (d decommission) Act(ctx context.Context, cluster *resource.Cluster) error
...
+		log.V(int(zapcore.WarnLevel)).Info("decommission statefulset does not have all replicas up")
 		return NotReadyErr{Err: errors.New("decommission statefulset does not have all replicas up")}
 	}
 
 	nodes := uint(cluster.Spec().Nodes)
 	//We do not scale down if the nodes field is less than 3
 	//TODO @alina add validation webhook (see https://github.com/cockroachdb/cockroach-operator/issues/245)
 	if nodes < 3 {
-		log.Info("We cannot decommission if there are less than 3 nodes", "nodes", nodes)
+		log.V(int(zapcore.WarnLevel)).Info("We cannot decommission if there are less than 3 nodes", "nodes", nodes)
```
keith-mcclellan, line 86: This should be Error not Warn

pkg/actor/deploy.go
```diff
@@ -126,12 +127,12 @@ func (d deploy) Act(ctx context.Context, cluster *resource.Cluster) error {
...
 		if changed {
-			log.Info("created/updated pdb, stopping request processing")
+			log.V(int(zapcore.DebugLevel)).Info("created/updated pdb, stopping request processing")
 			CancelLoop(ctx)
 			return nil
 		}
 	}
 
-	log.Info("completed")
+	log.V(int(zapcore.InfoLevel)).Info("completed")
```
keith-mcclellan, line 136: Completed what?  This message needs more context

pkg/actor/initialize.go
```diff
@@ -63,14 +64,14 @@ func (init initialize) Act(ctx context.Context, cluster *resource.Cluster) error
 	}
 	ss := &appsv1.StatefulSet{}
 	if err := init.client.Get(ctx, key, ss); err != nil {
-		log.Info("failed to fetch statefulset")
+		log.V(int(zapcore.WarnLevel)).Info("failed to fetch statefulset")
```
keith-mcclellan, line 67: This should be Error unless this is expected behavior for some reason

pkg/actor/initialize.go
```diff
@@ -95,7 +96,7 @@ func (init initialize) Act(ctx context.Context, cluster *resource.Cluster) error
 
 	cluster.SetFalse(api.NotInitializedCondition)
 
-	log.Info("completed")
+	log.V(int(zapcore.InfoLevel)).Info("completed")
```
keith-mcclellan, line 99: "completed" needs more context

pkg/actor/partitioned_update.go
```diff
@@ -95,7 +96,7 @@ func (up *partitionedUpdate) Act(ctx context.Context, cluster *resource.Cluster)
 
 	// nothing to be done
 	if dbContainer.Image == cluster.Spec().Image.Name {
-		log.Info("no version changes needed")
+		log.V(int(zapcore.DebugLevel)).Info("no version changes needed")
```
keith-mcclellan, line 99: this should be info

pkg/scale/drainer.go
```diff
@@ -171,7 +172,7 @@ func (d *CockroachNodeDrainer) makeDrainStatusChecker(id uint) func(ctx context.
 
 		isLive, replicasStr, isDecommissioning := record[8], record[9], record[10]
 
-		d.Logger.Info(
+		d.Logger.V(int(zapcore.DebugLevel)).Info(
```
keith-mcclellan, line 175: This should be info

pkg/update/update.go
```diff
@@ -268,7 +269,7 @@ func makeWaitUntilAllPodsReadyFuncInAllClusters(
 ) func(ctx context.Context, l logr.Logger) error {
 	return func(ctx context.Context, l logr.Logger) error {
 
-		l.Info("waiting until all pods are in the ready state")
+		l.V(int(zapcore.DebugLevel)).Info("waiting until all pods are in the ready state")
```
keith-mcclellan, line 272: This should be an info state

pkg/update/update.go
```diff
@@ -284,7 +285,7 @@ func makeWaitUntilAllPodsReadyFuncInAllClusters(
 				l.Error(err, "number of ready replicas is,  not equal to num CRDB pods")
 				return err
 			}
-			l.Info("all replicas are ready")
+			l.V(int(zapcore.DebugLevel)).Info("all replicas are ready")
```
keith-mcclellan, line 288: This should be an info state

pkg/update/update_cockroach_version.go
```diff
@@ -219,7 +220,7 @@ func kindAndCheckPreserveDowngradeSetting(
 		}
 		return s, nil
 	} else if isBackOneMajorVersion(wantVersion, currentVersion) {
-		l.Info("major rollback")
+		l.V(int(zapcore.DebugLevel)).Info("major rollback")
```
keith-mcclellan, line 223: This should be Error

pkg/update/update_cockroach_version.go
```diff
@@ -237,7 +238,7 @@ func kindAndCheckPreserveDowngradeSetting(
 		}
 		return s, nil
 	}
-	l.Info("unknown upgrade")
+	l.V(int(zapcore.WarnLevel)).Info("unknown upgrade")
```
keith-mcclellan, line 241: This should be error

pkg/update/update_cockroach_version_common.go
```diff
@@ -68,7 +69,7 @@ func makeIsCRBPodIsRunningNewVersionFunction(
 
 		crdbPod, err := clientset.CoreV1().Pods(stsNamespace).Get(update.ctx, podName, metav1.GetOptions{})
 		if k8sErrors.IsNotFound(err) { // this is not an error
-			l.Info("cannot find Pod", "podName", podName, "namespace", stsNamespace)
+			l.V(int(zapcore.DebugLevel)).Info("cannot find Pod", "podName", podName, "namespace", stsNamespace)
```
keith-mcclellan, line 72: This should be error

pkg/update/update_cockroach_version_common.go
```diff
@@ -85,19 +86,19 @@ func makeIsCRBPodIsRunningNewVersionFunction(
 
 				// TODO this is not an error but should return a wait status
 				if container.Image != cockroachImage {
-					l.Info("Pod is not updated to current image.")
+					l.V(int(zapcore.DebugLevel)).Info("Pod is not updated to current image.")
```
keith-mcclellan, line 89: This should be info

keith-mcclellan (changes requested):
There are a few log messages below that aren't discrete that need to be changed. Mostly the feedback is just on my opinion on the logging level for each of these messages.  @chrisseto and @alyshanjahani-crl may have additional feedback. Thanks!

pkg/actor/decommission.go
```diff
@@ -58,7 +59,7 @@ func (d decommission) Handles(conds []api.ClusterCondition) bool {
 
 func (d decommission) Act(ctx context.Context, cluster *resource.Cluster) error {
 	log := d.log.WithValues("CrdbCluster", cluster.ObjectKey())
-	log.Info("check decommission oportunities")
+	log.V(int(zapcore.InfoLevel)).Info("check decommission oportunities")
```
chrisseto, line 62: This seems like debug information, not info?

Rationale is that this log line is very similar to an "I AM HERE" message. It's only useful when trying to diagnose why the operator isn't doing what you expect it to be doing.

pkg/actor/deploy.go
```diff
@@ -46,7 +47,7 @@ func (d deploy) Handles(conds []api.ClusterCondition) bool {
 
 func (d deploy) Act(ctx context.Context, cluster *resource.Cluster) error {
 	log := d.log.WithValues("CrdbCluster", cluster.ObjectKey())
-	log.Info("reconciling resources")
+	log.V(int(zapcore.InfoLevel)).Info("reconciling resources")
```
chrisseto, line 50: Same debug comment as above.  We may want to make this message a bit more informative as well?

pkg/actor/request_cert.go
```diff
@@ -68,7 +69,7 @@ func (rc *requestCert) Act(ctx context.Context, cluster *resource.Cluster) error
 }
 
 func (rc *requestCert) issueNodeCert(ctx context.Context, log logr.Logger, cluster *resource.Cluster) error {
-	log.Info("requesting node certificate")
+	log.V(int(zapcore.InfoLevel)).Info("requesting node certificate")
```
chrisseto, line 72: Should this and the client certificate be the same logging level?

chrisseto (approved):
LGTM barring a few level adjustments.

Moving to zap proper seems like it would be a nice QoL improvement. 
Having to call `log.V(int(zapcore.InfoLevel))` for every log line is very tedious to type out.

My take away from the logging levels is this:
Debug - Useful information when diagnosing problems
Info - A concrete and notable action has been taken (just broad strokes though)
Warning - Something unexpected, but recoverable, has happened. Strictly speaking a missing STS would be recoverable if it was recreated or the error was caused by the inability to connect to the k8s API server.
Error - Something unexpected and unrecoverable has happened. Rejected certificates (Uncertain if a rejected cert can be re-approved), panics, and setup problems (inability to load credentials).

Does that sound correct? If so, should that be documented somewhere for future reference?

pkg/update/update.go
chrislovecnm: Why?

pkg/update/update_cockroach_version_common.go
chrislovecnm: No this is not an error.

pkg/update/update_cockroach_version_common.go
chrislovecnm: No

chrislovecnm: See https://github.com/cockroachdb/cockroach-operator/pull/323#pullrequestreview-597673974

pkg/update/update_cockroach_version_common.go
chrislovecnm: see the comment on line 70

pkg/update/update.go
keith-mcclellan: It was info previously, but because this isn't really a debug message so much as it's telling you that your replicas are healthy

keith-mcclellan (approved):
LGTM

PR #326, opened 2021-03-03
Generating Certificates

chrislovecnm:
# Why we are doing this

We are currently using the k8s certificate API to issue self-signed certs, and this K8s component is changing in future versions. The API does not allow an operator to issue a certificate with a proper self-signed CA. Also, OpenShift further restricts the use of the k8s certificate API. This PR uses the functionality in CRDB to generate self-signed certificates.

We do not recommend using self-signed certificates in production.

# This PR

This PR adds the capability to use the cockroach binary to generate self-signed TLS certificates to run a secure CRDB database. We are downloading the CRDB binary and adding the binary to the operator container. When the actor is enabled, the CRDB binary is executed and the CRDB binary generates certificates.  The actor is enabled via a feature flag and is currently disabled.

# Work not complete

- [x] Implement secret update
- [x] Test inside container
- [x] Test in cluster 
- [x] Error handling
- [x] Integration testing

# Future Work

In the future, we will replace this code with code that calls the code used in the CRDB code base.  That code is not usable and requires refactoring into another repository and a stand-alone library.


keith-mcclellan:
@alyshanjahani-crl We'd appreciate it if you reviewed this PR for us. Thanks!!!

pkg/actor/generate_cert.go
```diff
@@ -0,0 +1,242 @@
...
+		resource.NewKubeResource(ctx, rc.client, cluster.Namespace(), kube.DefaultPersister))
+	if client.IgnoreNotFound(err) != nil {
+		return errors.Wrap(err, "failed to get client TLS secret")
+	}
+
+	if secret.Ready() {
+		return nil
+	}
+
+	// This is replacing some code
```
alyshanjahani-crl, line 199: could you explain more in the comment? or perhaps move this to a constant.

pkg/security/certs.go
```diff
@@ -0,0 +1,336 @@
...
+		caType != ClientCAPem &&
+		caType != UICAPem {
+
+		return fmt.Errorf("caType argument to createCACertAndKey must be one of CAPem (%d), ClientCAPem (%d), or UICAPem (%d), got: %d", CAPem, ClientCAPem, UICAPem, caType)
+	}
+
+	certsDirParam := fmt.Sprintf(CERTS_DIR, certsDir)
+	caKeyParam := fmt.Sprintf(CA_KEY, caKeyPath)
+
+	// TODO do we need to move the certs to the correct path?
```
alyshanjahani-crl, line 140: does the command we're running in `execCmd` write the certs to a path that the generateCert rc specifics when calling ? 
If so i think we can ignore moving the certs to the correct path? Does the container get destroyed after the generated certs are persisted at k8s secrets?

pkg/security/certs.go
```diff
@@ -0,0 +1,336 @@
...
+	execCmd(args...)
+
+	//log.Infof(context.Background(), "generated node key: %s", keyPath)
+	return nil
+}
+
+// CreateUIPair creates a UI certificate and key using the UI CA.
+// The CA cert and key must load properly. If multiple certificates
+// exist in the CA cert, the first one is used.
+func CreateUIPair(certsDir, caKeyPath string, keySize int, lifetime time.Duration, overwrite bool, hosts []string) error {
```
alyshanjahani-crl, line 185: is this function to be removed? same with CreateClientPair below?

alyshanjahani-crl (approved):
Overall the changes lgtm, ill give an approval so im not gatekeeping the PR.
There's some commented out code that we should resolve before merging though.

I am still unfamiliar with alot of the public operator codebase, so early on my feedback may be limited. For example, through reviewing this pr: things like the bazel build, controlling feature flags, and this actor pattern w/ the reconciliation controllers 

Would anyone be up to let me shadow/pairProgram sometime ? It'd be a great help :)

alyshanjahani-crl:
Also do we have any integration/e2e tests of the public operator. Are they a part of this CI ?

pkg/security/certs.go
chrislovecnm: Yes, the code should be removed, but @keith-mcclellan do we need to generate the UIPair?

chrislovecnm:
@alyshanjahani-crl PTAL

keith-mcclellan:
@chrisseto would you mind taking a look at this please?

keith-mcclellan (approved):

pkg/security/certs.go
keith-mcclellan: We do need to generate a UIPair for certain install patterns, particularly when we move to the 3 port config. but we can merge as is and do that change when making the port change

PR #329, opened 2021-03-08
Updating dependencies and tools

chrislovecnm:
- updating k8s version to 1.20.2
- updating controller runtime version
- updating controller tools
- adding client generation
- updating all interfaces to comply with changes in the k8s client


pkg/testutil/env/sandbox.go
```diff
@@ -172,20 +175,16 @@ func createServiceAccount(s Sandbox) error {
 }
 
 func startCtrlMgr(t *testing.T, mgr manager.Manager) func() {
-	stop := make(chan struct{})
-	wg := &sync.WaitGroup{}
-	wg.Add(1)
+	ctx, cancel := context.WithCancel(context.Background())
```
chrislovecnm, line 178: @chrisseto this is the only way I could get this to work with the e2e tests.  Do you have a better way?

pkg/kube/helpers.go
```diff
@@ -193,15 +193,19 @@ func CreateOrUpdateAnnotated(ctx context.Context, c client.Client, obj runtime.O
...
-func mutate(f MutateFn, key client.ObjectKey, obj runtime.Object) error {
+func mutate(f MutateFn, key client.ObjectKey, obj client.Object) error {
 	if err := f(); err != nil {
 		return err
 	}
 
-	if newKey, err := client.ObjectKeyFromObject(obj); err != nil || key != newKey {
-		return fmt.Errorf("MutateFn cannot mutate object name and/or object namespace")
-	}
+	// TODO uncertain how to do this
```
chrisseto, line 201: There should be a helper floating around somewhere? It may be sufficient to just save the name & namespace and then compare them after calling `f`?

pkg/testutil/env/sandbox.go
```diff
@@ -81,30 +79,35 @@ type Sandbox struct {
...
-func (s Sandbox) setNamespaceIfMissing(obj apiruntime.Object) apiruntime.Object {
+func (s Sandbox) setNamespaceIfMissing(obj client.Object) client.Object {
 	accessor, _ := meta.Accessor(obj)
 	if accessor.GetNamespace() == "" {
-		obj = obj.DeepCopyObject()
-		accessor, _ = meta.Accessor(obj)
+		runtimeObject := obj.DeepCopyObject()
+		accessor, _ = meta.Accessor(runtimeObject)
 		accessor.SetNamespace(s.Namespace)
+		obj, ok := runtimeObject.(client.Object)
```
chrisseto, line 88: nit: I think you can just do `return runtimeObject.(client.Object)` and the runtime will panic for you if the cast fails.

pkg/testutil/env/sandbox.go
chrisseto: Can you elaborate on what you mean?

According to [the docs](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.8.3/pkg/manager#Manager) the preferred method of tearing down the manager is to cancel the passed in context.

If you wanted to remove the tear down func, you could use `t.Cleanup` to cancel the context? My only worry here is whether or not `ctrl-c` will interrupt the test. I think `go test` does some magic to make it work, but I'm not certain.

pkg/testutil/env/sandbox.go
```diff
@@ -172,20 +175,16 @@ func createServiceAccount(s Sandbox) error {
...
 
 	go func() {
-		defer wg.Done()
-		if err := mgr.Start(stop); err != nil {
+		if err := mgr.Start(ctx); err != nil {
 			t.Fatal(err)
 		}
 	}()
 
 	return func() {
```
chrisseto, line 186: nit: `return cancel` is a bit more clear

chrisseto (dismissed):
LGTM, just a handful of nits 👍

pkg/testutil/env/sandbox.go
chrislovecnm: Does the way I refactored it look correct?

e2e/e2e_test.go
```diff
@@ -139,8 +139,8 @@ func TestCreatesSecureCluster(t *testing.T) {
 			requireDatabaseToFunction(t, sb, builder)
 
 			//state, err := sb.Diff()
-			_, err := sb.Diff()
-			require.NoError(t, err)
+			//_, err := sb.Diff()
```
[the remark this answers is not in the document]
chrislovecnm: This is not going to work anymore without a major refactor.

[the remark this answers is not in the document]
chrislovecnm: @keith-mcclellan because we have a job now the e2e tests are not comparing exported yaml.  Personally, I do not like this type of testing, because it is very fragile.  Not certain we want to fix it.  But it is worth a conversation.

.bazelrc
```diff
@@ -1,7 +1,7 @@
 build --workspace_status_command hack/build/print-workspace-status.sh
-test --test_output=errors
+# test --test_output=errors
 #test --test_arg=-test.parallel=8
 
 # helpful for dev
-# test --test_output=streamed
-# test --test_arg=-test.v
+test --test_output=streamed
```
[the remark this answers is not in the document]
chrislovecnm: Need to back out changes

keith-mcclellan (approved):
LGTM

PR #333, opened 2021-03-10
Fixing version checker

chrislovecnm:
The version checker was implemented to only support versions
that are defined in the RELATED_IMAGES, but when using
image.name a user needs to have the capability to run any
version of CRDB.


pkg/actor/validate_version.go
```diff
@@ -73,11 +73,30 @@ func (v *versionChecker) Act(ctx context.Context, cluster *resource.Cluster) err
...
+	// If the image.name is set use that value and do not check that the
+	// version is set in the supported versions.
+	// If it is not set then pass through the if statement and check that
+	if cluster.Spec().Image.Name == "" {
+		if cluster.Spec().CockroachDBVersion == "" {
+			err := InvalidContainerVersionError{Err: errors.New("Cockroach image name and cockroachDBVersion api fields are not set, you must set one of them")}
+			log.Error(err, "invalid custom resources")
+			return err
+		}
+		log.V(int(zapcore.DebugLevel)).Info("User set CockroahDBVersion")
```
keith-mcclellan, line 86: typo

pkg/actor/validate_version.go
```diff
@@ -73,11 +73,30 @@ func (v *versionChecker) Act(ctx context.Context, cluster *resource.Cluster) err
...
+		// we check if the cockroachDBVersion is supported by the operator,
+		// this can return false only for api feild CockroachDBVersion
+		// The supported versions are set as enviroment variables in the operator manifest.
+		if !cluster.IsSupportedImage() {
+			err := InvalidContainerVersionError{Err: errors.New(fmt.Sprintf("crdb version %s not supported", cluster.Spec().CockroachDBVersion))}
+			log.Error(err, "The cockroackDBVersion API value is set to a value that is not supported by the operator. Supported versions are set via the RELATED_IMAGE env variables in the operator manifest.")
```
keith-mcclellan, line 92: typo

pkg/actor/validate_version.go
```diff
@@ -73,11 +73,30 @@ func (v *versionChecker) Act(ctx context.Context, cluster *resource.Cluster) err
...
+			return err
+		}
+		log.V(int(zapcore.DebugLevel)).Info(fmt.Sprintf("supported CockroahDBVersion %s", cluster.Spec().CockroachDBVersion))
+	} else {
+		log.V(int(zapcore.DebugLevel)).Info("User set image.name, using that field instead of cockroahDBVersion")
```
keith-mcclellan, line 97: typo

pkg/actor/validate_version.go
```diff
@@ -73,11 +73,30 @@ func (v *versionChecker) Act(ctx context.Context, cluster *resource.Cluster) err
...
+		log.V(int(zapcore.DebugLevel)).Info("User set CockroahDBVersion")
+		// we check if the cockroachDBVersion is supported by the operator,
+		// this can return false only for api feild CockroachDBVersion
+		// The supported versions are set as enviroment variables in the operator manifest.
+		if !cluster.IsSupportedImage() {
+			err := InvalidContainerVersionError{Err: errors.New(fmt.Sprintf("crdb version %s not supported", cluster.Spec().CockroachDBVersion))}
+			log.Error(err, "The cockroackDBVersion API value is set to a value that is not supported by the operator. Supported versions are set via the RELATED_IMAGE env variables in the operator manifest.")
+			return err
+		}
+		log.V(int(zapcore.DebugLevel)).Info(fmt.Sprintf("supported CockroahDBVersion %s", cluster.Spec().CockroachDBVersion))
```
keith-mcclellan, line 95: typo

pkg/resource/cluster.go
```diff
@@ -203,13 +203,18 @@ func (cluster Cluster) GetCockroachDBImageName() string {
 	supportedImages := getSupportedCrdbImages()
 	if cluster.Spec().CockroachDBVersion != "" {
 		if version, ok := cluster.LookupSupportedVersion(cluster.Spec().CockroachDBVersion); ok {
```
keith-mcclellan, line 205: It's possible I'm still missing something, but this logic doesn't look right to me.  The intent is to allow someone to either use one of the supported images via the CockroachDBVersion, or allow them to install any cockroachdb version.  In the latter case, we need to validate the version of crdb in the container and compare it to the current cluster version to make sure it is either a patch release of the current feature release, or the next newest feature release.  In the case of a new feature release, we need to also validate that any previous upgrade has been finalized.  This appears to kick out any image that isn't hosting an explicitly supported version of crdb.

keith-mcclellan (changes requested):
There were several typos that need to be fixed in the comments and log messages (which I tried to tag), and the version checker logic doesn't seem right to me still.  In addition, this is badly in need of some unit tests.

pkg/resource/cluster.go
alinadonisa: @keith-mcclellan  the CockroachDBVersion field can use the version checker only if it finds the image in the env var list we add. On line 203 we load the env vars we support and if this field is set we find a match on the supported versions to retrieve the container  image that will be loaded in the version checker job.   If we do not find an image we do not run  version checker job, the cluster will not be installed and we'll get an error on the status saying that we do not support the version given in the api field, or maybe to set the image field directly in the CR. With the api field we cannot run any version because we do not have the registry(docker, redhat etc), only with the image field we can do this logic at this time without hardcoding the registry.

chrislovecnm:
```log
PASS
Down(): stopin kind cluster...

I0311 21:57:37.288528  643387 app.go:87] ID for this run: "0eb286b1-085e-4ba0-8c71-a4b2a3f6c7c1"
Down(): deleting kind cluster...

Deleting cluster "test" ...
INFO: Elapsed time: 704.507s, Critical Path: 704.14s
INFO: 1 process: 1 processwrapper-sandbox.
INFO: Build completed successfully, 2 total actions
//e2e:go_default_test                                                    PASSED in 704.1s

INFO: Build completed successfully, 2 total actions
```

I am getting e2e passing on my build box.  There is a weird issue with request_certificate code, but since we are replacing that code I am not going to debug it.

keith-mcclellan:
This LGTM but I don't want to merge until @chrisseto also takes a look

PR #337, opened 2021-03-15
Fixing crdb binary in OSX

chrislovecnm:
I missed a path with downloading the crdb binary when running
OSX.  We should probably not have the rule the way it is, but it
is fine for now.


keith-mcclellan (approved):
LGTM

PR #340, opened 2021-03-15
Getting GKE E2E running and Bazel updates

chrislovecnm:
- updated docker bazel rules
- updated k8s bazel rules
- added a script to edit the manifest on the fly so that bazel works
- improving the capability to run GKE
- added make targets
- tested on GKE
- modified e2e testing to not start a cluster automatically
- modified e2e testing to accept a parameter for testing pvc resizing
- updated the kubetest2 builds in gcp

Once we have this in we will need to have the infra team have gcloud and the correct GCP permissions on the teamcity server in order to use this functionality on a nightly build.

TODO
- [x] update osx kubetest2 binaries

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -5,4 +5,4 @@ test --test_output=errors
 
 # helpful for dev
 # test --test_output=streamed
-# test --test_arg=-test.v
+test --test_arg=-test.v
--- a/Makefile
+++ b/Makefile
@@ -23,6 +23,9 @@ DOCKER_REGISTRY?=cockroachdb
 DOCKER_IMAGE_REPOSITORY?=cockroachdb-operator
 # Default bundle image tag
 APP_VERSION?=v1.6.12-rc.2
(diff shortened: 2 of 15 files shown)


keith-mcclellan:
@rail can we please get your eyes on this one? you're probably going to be more impacted by this than anyone as these tests will need to be run before we cut a release of the operator.

rail:
More tests is better! :)

I took a quick look at this. So far it looks straight forward. Regarding the release process. If we make these tests required either formally or in CI, the publishing part of the release should be the same.

Not sure if this quick look is qualified as an r+, but definitely an f+ (f for feedback).

chrislovecnm:
@keith-mcclellan can I get a LGTM?  I would like to use this code to test my new 3 port configuration.

keith-mcclellan:
LGTM

keith-mcclellan (approved):
LGTM

PR #345, opened 2021-03-16
Moving to 3 port database configuration

chrislovecnm:
- adding new API value SQLPort
- setting default SQLPort and moving default GRPCPort
- fixing unit tests
- updating start parameters for database


keith-mcclellan (approved):
LGTM

PR #350, opened 2021-03-17
Fixing missed SQL Ports

chrislovecnm:
Need to use the SQL Port instead of the GRPC Port to connect
to the database now.


keith-mcclellan (approved):
LGTM

PR #355, opened 2021-03-17
More work on getting integration testing working

chrislovecnm:
## Summary

- cleaning up Makefile
- downloading tester-exec
- added script to test if GKE has started
- above script needs to be replaced by go code, and work
for both GKE and kind
- get kind and gke working in same pattern

## TODO

- [x] replace bash with go
- [x] get kind working in the same pattern
- [x] update osx binaries

### DIFF
--- a/hack/bin/deps.bzl
+++ b/hack/bin/deps.bzl
@@ -28,6 +28,7 @@ def install():
     install_kubetest2()
     install_kubetest2_kind()
     install_kubetest2_gke()
+    install_kubetest2_exe()
     install_operator_sdk()
     install_kustomize()
     install_opm()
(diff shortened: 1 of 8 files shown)


hack/bin/deps.bzl
```diff
@@ -308,7 +309,23 @@ def install_kubetest2_gke():
...
     )
+## Fetch kubetest2-tester-exe binary used during e2e tests
+def install_kubetest2_exe():
+    # install kubetest2-exe binary
+    # TODO osx support
+    http_file(
+       name = "kubetest2_exe_darwin",
+       executable = 1,
+       sha256 = "a1cbe02f61931dbe6c8d1662442f42cb538c81e4ec8cdd40f548f0e05cbd55a7",
+       urls = ["https://storage.googleapis.com/crdb-bazel-artifacts/osx/kubetest2-tester-exec"],
```
chrislovecnm, line 320: TODO - need to upload osx

pkg/resource/cluster.go
```diff
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"os"
 	"strings"
+	"time"
```
chrislovecnm, line 23: Caught this running tests

keith-mcclellan (approved):
LGTM

PR #359, opened 2021-03-18
Cleaning up Makefile and adding documentation

chrislovecnm:
## Makefile updates

1. Removed old or targets we should not use
2. added a bunch of documentation on the e2e targets
3. moved a couple of targets around

# Testing updates
1. increased timeout on verify program
2. got pvc testing running on GKE

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,6 +47,12 @@ test/pkg:
 test/verify:
 	bazel test //hack/...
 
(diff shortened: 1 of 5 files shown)


keith-mcclellan (approved):
LGTM

rail (commented):
LGTM, :shipit:

Issue #363, opened 2021-03-19
Consider using GitHub Actions for CI


rail:
Pros:
* Better visibility (no TeamCity access required to see the logs)
* Self serve
* CI configs are versioned and don't need to be synced with the code changes

Cons:
* We may need to run some sensitive tasks (nightlies and releases) in TeamCity, so we need to support 2 CI systems

Any other thoughts?

keith-mcclellan:
I would LOVE to move this to github actions. It will greatly help us encourage community contributions to the operator.

keith-mcclellan:
@chrislovecnm thoughts?

chrislovecnm:
@keith-mcclellan I have no experience with GitHub actions.  So I will let you and @rail decide.

rail:
Thank yo. I'll merge it later today unless I hear any objections.

rail:
I mean I will enable GitHub actions, but won't make it mandatory yet.

rail:
Done

PR #371, opened 2021-03-19
Add GitHub Actions support

rail:
The patch reflects the CI builds we have in TeamCity.

* Run against bazel 3.7.2 and 4.0.0

It will be running in parallel with TeamCity for now and won't be blocking merges. We can make it mandatory after we decide in #363

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,0 +1,62 @@
+# Copyright 2021 The Cockroach Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: CI
+
+on:
+  # Triggers the workflow on push or pull request events but only for the master branch
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+jobs:
+  e2e-kind:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        bazel: ["4.0.0", "3.7.2"]
+    steps:
+      - uses: actions/checkout@v2
+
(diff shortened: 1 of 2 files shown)


chrislovecnm:
We tried to run kind?

rail:
> Have we tried to run kind?

Yeah, it worked fine in my fork.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,59 @@
+#
```
chrislovecnm, line 1: This is the wrong file header and should not pass CI. See the boilerplates directory.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,59 @@
...
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Setup Bazel
+        uses: abhinavsingh/setup-bazel@v3
+        with:
+          version: ${{ matrix.bazel }}
+
+      - name: Tests
+        run: bazel test //apis/... //pkg/... //hack/... --test_output=streamed --test_output=errors
```
chrislovecnm, line 58: Should we run the make target? That way we are all running the same thing

chrislovecnm (commented):
Couple of questions

.github/workflows/ci.yaml
rail: Wrong copy/paste, doh.

.github/workflows/ci.yaml
rail: I copied it from TeamCity. Looks like `test/all` is missing `//hack/...`. Shall we add it or use another target? Or multiple targets?

chrislovecnm: Add it to that target

rail: Done

rail:
Opps, my editor removed all the trailing spaces from `Makefile`. I can revert those if you care.

chrislovecnm:
@rail love cleanup :)

chrislovecnm (dismissed):
LGTM

chrislovecnm:
@keith-mcclellan can you take a look?

rail:
I pushed the latest change to my fork's master here: https://github.com/rail/cockroach-operator/actions/runs/668802635

The e2e test is flaky often, but it passed a few times. Since this CI is not blocking yet, I think it's ok to have it that way now and fix as we go.

rail:
It may be actually being killed because it takes too long. I wonder if we can split the e2e tests and run in paralllel.

rail:
My force push cleaned the r+, oops

chrislovecnm:
@keith-mcclellan PTAL - @rail someone from CRL has to review the PRs

keith-mcclellan (approved):
LGTM

PR #377, opened 2021-03-24
add  annotation on cluster for cert request expire timestamp

alinadonisa:
## Synopsis

The purpose of this PR is to provide a mechanism to manually rotate certificates. First step will be to signal the user the expiration date of the node certificate, this way the user can acknowledge that he/she needs to manually replace the existing certificate and restart the cluster afterwards.

We are taking in consideration 2 types of restart:

- **FullCluster** restart that will be use when CA certificate is replaced
- **Rolling** restart used in the other cases

For future development we are taking in consideration also to add integration with certmanager.

## Work completed this week

- I added implementation for Rolling restart and FullCluster restart
- I fixed error from sync version checker and request cert action 
- I worked on refactory  the code for version checker 

## Work pending:

-  Restart runs only once now, we should think of a method to run it once and not interfere with the operator workflow.


pkg/resource/cluster.go
```diff
@@ -204,6 +205,12 @@ func (cluster Cluster) SetAnnotationContainerImage(containerimage string) {
 	}
 	cluster.cr.Annotations[CrdbContainerImageAnnotation] = containerimage
 }
+func (cluster Cluster) SetAnnotationCAExpiration(version string) {
```
chrislovecnm, line 208: Need godocs please

pkg/resource/cluster.go
```diff
@@ -38,6 +38,7 @@ const (
 	CrdbContainerImageAnnotation = "crdb.io/containerimage"
 	CrdbVersionAnnotation        = "crdb.io/version"
 	CrdbHistoryAnnotation        = "crdb.io/history"
+	CrdbCAExpirationAnnotation   = "crdb.io/certexpiration"
```
chrislovecnm, line 41: One thing I noticed is that crdb.io does not exist. @keith-mcclellan, which domain do we want on these?

chrislovecnm (commented):
Just a couple of comments and please check on CI

pkg/resource/cluster.go
keith-mcclellan: why do we have a url here at all @chrislovecnm ?

pkg/actor/cluster_restart.go
```diff
@@ -0,0 +1,63 @@
...
+
+	config *rest.Config
+}
+
+//GetActionType returns api.VersionCheckerAction action used to set the cluster status errors
+func (v *clusterRestart) GetActionType() api.ActionType {
+	return api.ClusterRestartAction
+}
+
+func (v *clusterRestart) Handles(conds []api.ClusterCondition) bool {
```
chrislovecnm, line 53: Godocs please

pkg/resource/cluster.go
```diff
@@ -204,6 +205,12 @@ func (cluster Cluster) SetAnnotationContainerImage(containerimage string) {
 	}
 	cluster.cr.Annotations[CrdbContainerImageAnnotation] = containerimage
 }
+func (cluster Cluster) SetAnnotationCertExpiration(certExpiration string) {
```
chrislovecnm, line 208: Godocs

keith-mcclellan (changes requested):
In general I think this does what we want - we can merge once you resolve Chris's comments.  In the future, I'd have preferred that a feature like ClusterRestart be done in it's own PR because there are other places to use that feature than just rotating the certs and we want to make sure the implementation is general enough.  It is much easier to review one feature at a time than two together.  Thanks!!!

pkg/resource/cluster.go
alinadonisa: @keith-mcclellan if the prefix is omitted, the annotation Key is presumed to be private to the user. Automated system components (e.g. kube-scheduler, kube-controller-manager, kube-apiserver, kubectl, or other third-party automation) which add annotations to end-user objects must specify a prefix. Please see k8s documentation for annotations: https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/

chrislovecnm:
@alinadonisa we need some godocs updates on this code please, and then we can merge

Makefile
```diff
@@ -18,13 +18,13 @@
 #
 
 # values used in workspace-status.sh
-DOCKER_REGISTRY?=cockroachdb
+DOCKER_REGISTRY?=us.gcr.io/chris-love-operator-playground
```
chrislovecnm, line 21: Need to roll this back please

manifests/operator.yaml
```diff
@@ -257,16 +257,16 @@ spec:
       serviceAccountName: cockroach-operator-sa
       containers:
         - name: cockroach-operator
-          image: cockroachdb/cockroach-operator:v1.6.12-rc.2
+          image: us.gcr.io/chris-love-operator-playground/cockroachdb-operator:v1.9.17
```
chrislovecnm, line 260: Back out changes please

pkg/actor/cluster_restart.go
```diff
@@ -0,0 +1,200 @@
...
+
+	config *rest.Config
+}
+
+//GetActionType returns api.ClusterRestartAction action used to set the cluster status errors
+func (r *clusterRestart) GetActionType() api.ActionType {
+	return api.ClusterRestartAction
+}
+
+func (r *clusterRestart) Handles(conds []api.ClusterCondition) bool {
```
chrislovecnm, line 64: GoDocs please

pkg/actor/cluster_restart.go
```diff
@@ -0,0 +1,200 @@
...
+	return utilfeature.DefaultMutableFeatureGate.Enabled(features.ClusterRestart) && condition.True(api.InitializedCondition, conds) || condition.False(api.InitializedCondition, conds)
+}
+
+func (r *clusterRestart) Act(ctx context.Context, cluster *resource.Cluster) error {
+	log := r.log.WithValues("CrdbCluster", cluster.ObjectKey())
+	log.V(int(zapcore.DebugLevel)).Info("starting cluster restart action")
```
chrislovecnm, line 70: Please use the log package I wrote

pkg/actor/cluster_restart.go
```diff
@@ -0,0 +1,200 @@
...
+	if cluster.Spec().RestartType == api.RollingRestart {
+		log.V(int(zapcore.DebugLevel)).Info("initiating rolling restart action")
+		// TODO statefulSetIsUpdating is not quite working as expected.
+		// I had to check status.  We should look at the update code in partition update to address this
+		if statefulSetIsUpdating(statefulSet) {
+			return NotReadyErr{Err: errors.New("restart statefulset is updating, waiting for the update to finish")}
+		}
+		status := &statefulSet.Status
+		if status.CurrentReplicas == 0 || status.CurrentReplicas < status.Replicas {
+			log.Info("restart statefulset does not have all replicas up")
```
chrislovecnm, line 98: Use logging levels please

pkg/actor/cluster_restart.go
```diff
@@ -0,0 +1,200 @@
...
+			return NotReadyErr{Err: errors.New("restart cluster statefulset does not have all replicas up")}
+		}
+		timeNow := metav1.Now()
+		// stsName := statefulSet.Name
```
chrislovecnm, line 102: Remove dead code

pkg/actor/cluster_restart.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+	return update.RollingRestart(ctx, updateRoach, k8sCluster, r.log)
+}
+
+//ScaleSts updates the replicas of the statefullset to the value from parameters
+//and waits until the statefullset has currentreplicas equal with the desired replicas
+//We will use this in FullRestart logic, by setting replicas to 0 and
+//afterwords setting to the original number of nodes, this way the cluster
+//will load the new CA certs
+func (r *clusterRestart) ScaleSts(ctx context.Context, sts *appsv1.StatefulSet, l logr.Logger, clientset *kubernetes.Clientset, replicas int32) error {
```
chrislovecnm, line 176: Does this need to be public func?

chrislovecnm (commented):
Nitpicks mostly and we need a test

apis/v1alpha1/cluster_types.go
```diff
@@ -97,6 +97,13 @@ type CrdbClusterSpec struct {
 	// Default: ""
 	// +optional
 	CockroachDBVersion string `json:"cockroachDBVersion,omitempty"`
+	// (Optional) RestartType if set with value FullCluster restarts the entire cluster.
+	// FullCluster option is used when the CA cert is replaced and we need to restart all nodes at once.
+	// If the value is set to Rollback it will restart one node at a time.
```
udnay, line 102: /typo? `Rollback` should be `Rolling`?

config/crd/bases/crdb.cockroachlabs.com_crdbclusters.yaml
```diff
@@ -260,6 +260,9 @@ spec:
                     description: 'Requests describes the minimum amount of compute resources required. If Requests is omitted for a container, it defaults to Limits if that is explicitly specified, otherwise to an implementation-defined value. More info: https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/'
                     type: object
                 type: object
+              restartType:
+                description: '(Optional) RestartType if set with value FullCluster restarts the entire cluster. FullCluster option is used when the CA cert is replaced and we need to restart all nodes at once. If the value is set to Rollback it will restart one node at a time. Other values are invalid. Default: ""'
```
udnay, line 264: Same comment as above

pkg/resource/cluster.go
keith-mcclellan: got it thanks

keith-mcclellan (approved):
LGTM

PR #378, opened 2021-03-24
Move version to standalone file

rail:
I would appreciate your feedback on this change. 

One question that I have is wether we should keep the generated metadata commited in the repo or  we generate it runtime as a part of the release process and discard the changes.

Before this change in order to bump the operator version one would need
to edit `Makefile` and `manifests/operator.yaml` to set the operator
version. Also the OpenShift metadata bundle version has to be set and
someone has to run a command to generate a new directory with metadata
files.

This patch moves the version into a separate file to make it more
automation friendly. At the same time it starts using the same version
(except the "v" prefix) for OpenShift metadata bundle.

Added a make target that automates a regular release preparation step.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cockroach-operator/378)


Makefile
```diff
@@ -174,6 +175,16 @@ dev/syncdeps:
 #
 # Release targets
 #
+
+.PHONY: release/versionbump
+release/versionbump:
+	$(MAKE) CHANNEL=stable IS_DEFAULT_CHANNEL=1 release/update-pkg-manifest && \
```
chrislovecnm, line 181: @alinadonisa correct me if I am wrong but I think we don’t release to stable right away.

Makefile
```diff
@@ -189,11 +200,11 @@ release/image:
 #RED HAT IMAGE BUNDLE
 RH_BUNDLE_REGISTRY?=registry.connect.redhat.com/cockroachdb
 RH_BUNDLE_IMAGE_REPOSITORY?=cockroachdb-operator-bundle
-RH_BUNDLE_VERSION?=1.2.5
+RH_BUNDLE_VERSION?=$(VERSION)
 RH_DEPLOY_PATH="deploy/certified-metadata-bundle"
 RH_DEPLOY_FULL_PATH="$(RH_DEPLOY_PATH)/cockroach-operator/"
-RH_COCKROACH_DATABASE_IMAGE=registry.connect.redhat.com/cockroachdb/cockroach:v20.2.5
-RH_OPERATOR_IMAGE?=registry.connect.redhat.com/cockroachdb/cockroachdb-operator:v1.6.12-rc.1
+RH_COCKROACH_DATABASE_IMAGE=registry.connect.redhat.com/cockroachdb/cockroach:$(COCKROACH_DATABASE_VERSION)
```
chrislovecnm, line 206: We are going to have to push all of the versions in the operator.yaml - @keith-mcclellan comments?

chrislovecnm (commented):
Just a couple of recommendations. Looks great! Did we want to do this with bazel instead of make? Just a thought

Makefile
alinadonisa: > @alinadonisa correct me if I am wrong but I think we don’t release to stable right away.

@rail, @Chris we will release on beta channel first and after the feedback  we go for the stable channel

alinadonisa:
> One question that I have is wether we should keep the generated metadata commited in the repo or we generate it runtime as a part of the release process and discard the changes.

@rail  we should keep the generated files that are **published** on OpenShift on git.  If we could automate this behaviour it would be great. For now it is enough to use the generated files from git.

Please take in consideration that if the CI fails we will need to increment the RH_BUNDLE_VERSION, regenerate the metadata and push again.
 
> Before this change in order to bump the operator version one would need
> to edit `Makefile` and `manifests/operator.yaml` to set the operator
> version. Also the OpenShift metadata bundle version has to be set and
> someone has to run a command to generate a new directory with metadata
> files.
> 
> This patch moves the version into a separate file to make it more
> automation friendly. At the same time it starts using the same version
> (except the "v" prefix) for OpenShift metadata bundle.
> 
> Added a make target that automates a regular release preparation step.
> 
> This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cockroach-operator/378)



> I would appreciate your feedback on this change.
> 
> One question that I have is wether we should keep the generated metadata commited in the repo or we generate it runtime as a part of the release process and discard the changes.
> 
> Before this change in order to bump the operator version one would need
> to edit `Makefile` and `manifests/operator.yaml` to set the operator
> version. Also the OpenShift metadata bundle version has to be set and
> someone has to run a command to generate a new directory with metadata
> files.
> 
> This patch moves the version into a separate file to make it more
> automation friendly. At the same time it starts using the same version
> (except the "v" prefix) for OpenShift metadata bundle.
> 
> Added a make target that automates a regular release preparation step.
> 
> This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cockroach-operator/378)

Makefile
keith-mcclellan: So if we're pushing a version of the operator container, we decided we're always going to push a metadata bundle even if nothing has changed and vice versa.  Otherwise it'll get impossible to track the inter-dependency.

rail (commented):
> Please take in consideration that if the CI fails we will need to increment the RH_BUNDLE_VERSION, regenerate the metadata and push again.
I believe with the proposed approach because we use the same version for the operator and the bundle, this should be less an issue.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cockroach-operator/378#-:-MWyLgLb5p0fnLkFbDSw:b-f75efy)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @alinadonisa, @chris, and @chrislovecnm)

___
*[Makefile, line 181 at r1](https://reviewable.io/reviews/cockroachdb/cockroach-operator/378#-MWeL7enBJegRNB24ewK-r1-181:-MWyLyxmDivqMO8Yxy7G:b-tb02r1) ([raw file](https://github.com/cockroachdb/cockroach-operator/blob/9259d956b09d40efce23e6ab0def4a35a162cf16/Makefile#L181)):*
<details><summary><i>Previously, alinadonisa (Alina) wrote…</i></summary><blockquote>

> @alinadonisa correct me if I am wrong but I think we don’t release to stable right away.

@rail, @Chris we will release on beta channel first and after the feedback  we go for the stable channel 
</blockquote></details>

I'll set the default values in `Makefile` to use beta and add some instructions in `README.md`.

keith-mcclellan:
@rail Chris made a bunch of changes to the makefile to resolve a bunch of build issues and move us to bazel 4, can you update this? Happy to give you my feedback at that point.

rail:
Rebased.

keith-mcclellan (approved):
LGTM

PR #384, opened 2021-03-29
Fixing K8S Cluster Stamping

chrislovecnm:
This PR addresses the updating of --stamp variables
via moving the stamp variables to "STABLE" values.
Moving the values to "STABLE" seems to force the updating
of rules_k8s files better.

Secondly, since this is working I fixed applying the GKE manifest
during the e2e cluster.

Lastly I added setting a missing environment variable to the apply
commands when running rules_k8s bazel commands.

### DIFF
--- a/hack/build/print-workspace-status.sh
+++ b/hack/build/print-workspace-status.sh
@@ -50,9 +50,9 @@ RH_DEPLOY_PATH ${RH_DEPLOY_PATH:-deploy/certified-metadata-bundle/cockroach-oper
 RH_BUNDLE_VERSION ${RH_BUNDLE_VERSION:-""}
 RH_BUNDLE_IMAGE_TAG ${RH_BUNDLE_IMAGE_TAG:-${RH_BUNDLE_VERSION:-""}}
   
-IMAGE_REGISTRY ${DEV_REGISTRY:-us.gcr.io/chris-love-operator-playground}
+STABLE_IMAGE_REGISTRY ${DEV_REGISTRY:-us.gcr.io/chris-love-operator-playground}
 
-CLUSTER ${K8S_CLUSTER:-gke_chris-love-operator-playground_us-central1-a_test}
+STABLE_CLUSTER ${K8S_CLUSTER:-gke_chris-love-operator-playground_us-central1-a_test}
 NUMBER_COMMITS_ON_BRANCH $(git rev-list $(git rev-parse --abbrev-ref HEAD) | wc -l)
 
 gitCommit ${KUBE_GIT_COMMIT-}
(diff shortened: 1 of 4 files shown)


hack/build/print-workspace-status.sh
```diff
@@ -50,9 +50,9 @@ RH_DEPLOY_PATH ${RH_DEPLOY_PATH:-deploy/certified-metadata-bundle/cockroach-oper
 RH_BUNDLE_VERSION ${RH_BUNDLE_VERSION:-""}
 RH_BUNDLE_IMAGE_TAG ${RH_BUNDLE_IMAGE_TAG:-${RH_BUNDLE_VERSION:-""}}
   
-IMAGE_REGISTRY ${DEV_REGISTRY:-us.gcr.io/chris-love-operator-playground}
+STABLE_IMAGE_REGISTRY ${DEV_REGISTRY:-us.gcr.io/chris-love-operator-playground}
```
keith-mcclellan, line 53: can we not hardcode a value in here please?  we're going to need to clean this up soon anyway, might as well get started if we're touching this file

hack/build/print-workspace-status.sh
```diff
@@ -50,9 +50,9 @@ RH_DEPLOY_PATH ${RH_DEPLOY_PATH:-deploy/certified-metadata-bundle/cockroach-oper
...
 
-CLUSTER ${K8S_CLUSTER:-gke_chris-love-operator-playground_us-central1-a_test}
+STABLE_CLUSTER ${K8S_CLUSTER:-gke_chris-love-operator-playground_us-central1-a_test}
```
keith-mcclellan, line 55: same as line 53, let's pass this value into the script rather than having it default to one dev's gcr

keith-mcclellan (changes requested):
This looks fine but I'd prefer to generalize the hack script a bit more, don't like it being defaulted to one dev's repo

hack/build/print-workspace-status.sh
chrislovecnm: It is not harder coded but we have a default value. We are passing this value in with the Makefile, but we don’t  always use the Makefile. What would you like?

chrislovecnm: If we don’t default if I probably need to add error handling. And these are the defaults that the developer use to make the scripts easier for them.

chrislovecnm: There are other defaults that we may want to clean up as well, and defaults in the Makefile. Without default values it does add a layer of complexity btw.

https://github.com/cockroachdb/cockroach-operator/blob/c3914639b6f9dc1685c8cc99e0b478d0e491d077/hack/build/print-workspace-status.sh#L42

keith-mcclellan (approved):
LGTM - opened an issue to clean up the defaults separately

PR #400, opened 2021-04-05
Updating UBI image with yum commands

chrislovecnm:
Building a tar layer from running YUM update commands
using the UBI image. We now use that tar ball as part
of the package.


cmd/cockroach-operator/BUILD.bazel
```diff
@@ -2,6 +2,10 @@ load("@io_bazel_rules_go//go:def.bzl", "go_binary", "go_library")
 load("@io_bazel_rules_docker//container:container.bzl", "container_image")
 load("@io_bazel_rules_docker//go:image.bzl", "go_image")
 load("@bazel_tools//tools/build_defs/pkg:pkg.bzl", "pkg_tar")
+load("@io_bazel_rules_docker//container:container.bzl", "container_image")
```
alinadonisa: this is duplicate

alinadonisa (changes requested):

alinadonisa (approved):

alyshanjahani-crl (approved):

PR #408, opened 2021-04-13
Use annotation instead of restartType api field

alinadonisa:
## Synopsis

The purpose of this PR is to provide a mechanism to manually rotate certificates. First step will be to signal the user the expiration date of the node certificate, this way the user can acknowledge that he/she needs to manually replace the existing certificate and restart the cluster afterwards.

We are taking in consideration 2 types of restart:

- **FullCluster** restart that will be use when CA certificate is replaced
- **Rolling** restart used in the other cases

For future development we are taking in consideration also to add integration with certmanager.

## Work completed this week

- I have replaced the api field with an annotation
- I added also CA key in the secret

## Work pending:

-  Testing  and bugfixing :)


keith-mcclellan (commented):
Functionally this LGTM but the e2e tests are still failing.  @chrislovecnm please also review before I approve since I can't validate via the e2e tests

keith-mcclellan (approved):
LGTM

PR #372, opened 2021-03-22
Upgrades patch

alinadonisa:
## Synopsis

The purpose of this PR is to add the latest fixes  on the refactoring of the mechanism already implemented and described in the issue https://github.com/cockroachdb/cockroach-operator/issues/200.

## Work completed this week

- I made a fix on maxlenght for job name see PR https://github.com/cockroachdb/cockroach-operator/pull/358
- I added some changes for the job name unique task
- I changed for the version checker to use the pod from the latest job  on each run
- I added the IsImagePullBackOff function used not to requeue the request 

## Work pending:
-  Some border cases to be discussed with Keith (like the failing job should be manually deleted when a new cluster is created)
-  Bazel builds needs to be checked and documentation should be updated 
-  Cert renew certificate implementation
-  Finish release automation for Open Shift(dev and certified)

PR #358, opened 2021-03-18
add maxlength on jobName using slug

alinadonisa:

Issue #200, opened 2020-10-26
CockroachDB Image name issue (placeholder)


johnrk:
We have a hacky way of handling cockroachdb image name

the code expects the image to be named cockroach/cockroach

chrislovecnm:
Talking with @keith-mcclellan we need to also support SHAs as well.  We need to include a CRD field for the version of the database, and then apply that as a label or an annotation to the statefulset.  When the CRD and the statefulset are out of sync, then we update the pods.

chrislovecnm:
@johnrk this may block openshift CI btw.

chrislovecnm:
Here is the design that @chrisseto @alinadonisa and I worked out

1. The user defines the image in the CR, which already exists.  This is not a change in the code.  The next steps are code changes.
2. The operator determines that the version of the image differs from the image in the statefulset.
3. The operator creates a new Job or a Pod with the image.  This creates a new CR database.
4. The operator determines the version of the CR binary through one of two methods.  @chrisseto is asking DB gurus for the preferred method. We can run a job and parse the CLI output, or we can start a database and make a SQL call to get the binary/DB version.
5. If the database cannot upgrade because of version mismatches, the operator writes the status and does not upgrade.
6. If the database can upgrade, we upgrade with the current code.

Here is what we need

1. We need status working and completed.  The operator needs to have a status field to store the information that an error has occurred where it cannot upgrade or downgrade a database.  Otherwise, we will have recurring loops.
2. We need an in-memory database to store image version and DB version hash
3. We need to write the code to do steps 2-6

keith-mcclellan:
The openshift marketplace requires that we pass in the CRDB image via an environment variable (RELATED_IMAGES_*) - will this support that? See https://redhat-connect.gitbook.io/certified-operator-guide/appendix/offline-enabled-operators#all-operators-helm-ansible-or-golang

chrislovecnm:
@keith-mcclellan I pinged Zach about the RELATED_IMAGES env variable.  I need to understand how that variable is used during testing.  We cannot just upgrade an existing database depending on that variable.  The OpenShift pattern is flawed when it comes to this.  I need more technical information to work out a design with @chrisseto and @alinadonisa.

chrislovecnm:
@alinadonisa is this supported yet? https://github.com/operator-framework/operator-lifecycle-manager/blob/master/doc/contributors/design-proposals/related-images.md

alinadonisa:
[...]

PR #415, opened 2021-04-15
Upgrades patch

chrislovecnm:
Redoing #372 

## Synopsis

The purpose of this PR is to add the latest fixes  on the refactoring of the mechanism already implemented and described in the issue https://github.com/cockroachdb/cockroach-operator/issues/200.

## Work completed this week

- I made a fix on maxlenght for job name see PR https://github.com/cockroachdb/cockroach-operator/pull/358
- I added some changes for the job name unique task
- I changed for the version checker to use the pod from the latest job  on each run
- I added the IsImagePullBackOff function used not to requeue the request 

## Work pending:
-  Some border cases to be discussed with Keith (like the failing job should be manually deleted when a new cluster is created)
-  Bazel builds needs to be checked and documentation should be updated 
-  Cert renew certificate implementation
-  Finish release automation for Open Shift(dev and certified)


Makefile
```diff
@@ -22,6 +22,8 @@ DOCKER_REGISTRY?=cockroachdb
 DOCKER_IMAGE_REPOSITORY?=cockroachdb-operator
 VERSION?=$(shell cat version.txt)
 APP_VERSION?=v$(VERSION)
+# Default bundle image tag
+APP_VERSION?=v1.6.12-rc.2
```
rail, line 26: This redefines the version coming from `version.txt`. Can we avoid doing this?

keith-mcclellan (approved):
LGTM

rail (changes requested):

rail (approved):

PR #423, opened 2021-04-20
Clean up bazel cache for release builds

rail:
When the `release/image` target is run multiple times, it doesn't
refresh the base RedHat UBI image. As a result we may end up with an
image without security updates.

This change ensures that we run `bazel clean --expunge` for the release
target.


chrisseto (approved):
I'm not a bazel expert but LGTM!

PR #429, opened 2021-04-21
Update all ubi packages

rail:
The current command doesn't update the image to the state when it passes
the RedHat Connect checks.

### DIFF
--- a/cmd/cockroach-operator/BUILD.bazel
+++ b/cmd/cockroach-operator/BUILD.bazel
@@ -38,7 +38,7 @@ container_run_and_commit_layer(
     name = "ubi_update",
     commands = [
         "microdnf install yum",
-        "yum -y update-minimal --security --sec-severity=Important --sec-severity=Critical",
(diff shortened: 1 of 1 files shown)


keith-mcclellan (approved):
LGTM

cmd/cockroach-operator/BUILD.bazel
```diff
@@ -38,7 +38,7 @@ container_run_and_commit_layer(
     name = "ubi_update",
     commands = [
         "microdnf install yum",
-        "yum -y update-minimal --security --sec-severity=Important --sec-severity=Critical",
```
chrislovecnm, line 41: Do we need the -v?

rail: Not really, but it may help debugging in the future. 🤷

keith-mcclellan: Probably not but it doesn't hurt, just gives the verbose output

PR #432, opened 2021-04-21
Add "latest" tag

rail:
In some cases the operator image should be also pushed to the `latest`
tag.`

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -27,6 +27,15 @@ container_push(
 )
 
 container_push(
+    name = "push_operator_image_latest",
+    format = "Docker",
+    image = "//cmd/cockroach-operator:operator_image",
+    registry = "{STABLE_DOCKER_REGISTRY}",
+    repository = "{STABLE_IMAGE_REPOSITORY}",
+    tag = "latest",
+)
+
+container_push(
     name = "push_operator_bundle_image",
     format = "Docker",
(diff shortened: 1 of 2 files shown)


chrislovecnm:
@rail in what cases? I have lost production pods because of the wonderful “latest” tag.

rail:
> @rail in what cases? I have lost production pods because of the wonderful “latest” tag. So I am not a huge fan of having latest. Do we need this?

I totally agree, "latest" is evil. Right now we publish to dockerhub and there is the "latest" tag. If we are OK removing it, I would retract this PR. :)

rail:
@keith-mcclellan what you say, do we really need the "latest" tag in docker hub?

keith-mcclellan:
We either need to keep the 'latest' tag up to date or remove it entirely. I don't think DockerHub allows you to remove it.  I think the Red Hat registry also requires the latest tag.  If we can remove them entirely I'm all for it.

keith-mcclellan:
Screen Shot 2021-04-22 at 10 40 57 AM
The Red Hat Registry auto-populates latest.

keith-mcclellan:
fwiw our configuration files should NEVER reference the latest tag to avoid the scenario that @chrislovecnm is describing because I agree it's not something we want to encourage people to use.

rail:
I'll remove the latest tag from dockerhub.

PR #458, opened 2021-04-28
Capability to set environment variables and telemetry

chrislovecnm:
This PR does three things.

1. Any env prefixed with CRDB that is in the operator is now
passed into the pods.
2. Telemetry now adds the dynamic value of the k8s distro that
it is running on.
3. Adds a new CRD field that allows the user to set one of more Pod environment variables.


keith-mcclellan (approved):
LGTM - please backlog an issue to enhance e2e to make sure we validate these variables.

PR #477, opened 2021-04-29
StatefulSet Changes

chrislovecnm:
- Dynamically set cache and max sql memory if the user does
not set a CRD value
- sts is set to Parallel pod management

### DIFF
--- a/apis/v1alpha1/cluster_defaults.go
+++ b/apis/v1alpha1/cluster_defaults.go
@@ -46,14 +46,6 @@ func SetClusterSpecDefaults(cs *CrdbClusterSpec) {
 		cs.MaxUnavailable = &DefaultMaxUnavailable
 	}
 
-	if cs.Cache == "" {
-		cs.Cache = "25%"
-	}
-
-	if cs.MaxSQLMemory == "" {
-		cs.MaxSQLMemory = "25%"
-	}
-
 	if cs.Image.PullPolicyName == nil {
 		policy := v1.PullIfNotPresent
 		cs.Image.PullPolicyName = &policy
--- a/pkg/kube/helpers.go
+++ b/pkg/kube/helpers.go
@@ -25,6 +25,8 @@ import (
 	"github.com/pkg/errors"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
+	rbac "k8s.io/api/rbac/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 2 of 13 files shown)


keith-mcclellan (approved):
LGTM

apis/v1alpha1/cluster_defaults.go
```diff
@@ -47,11 +47,11 @@ func SetClusterSpecDefaults(cs *CrdbClusterSpec) {
 	}
 
 	if cs.Cache == "" {
-		cs.Cache = "25%"
+		cs.Cache = "$(expr $MEMORY_LIMIT_MIB / 4)MiB"
 	}
 
 	if cs.MaxSQLMemory == "" {
-		cs.MaxSQLMemory = "25%"
+		cs.MaxSQLMemory = "$(expr $MEMORY_LIMIT_MIB / 4)MiB"
```
chrislovecnm, line 54: @chrisseto this is not working, I think, because we do not run an exec command. We cannot run an exec command because it does not work on openshift. Any ideas?

keith-mcclellan: we are actually running an exec in the container - the entrypoint script does that (see [cockroach.sh](https://github.com/cockroachdb/cockroach/blob/master/build/deploy/cockroach.sh)).  We can't explicitly do it outside of a bash script is all.

keith-mcclellan: Here are a couple of other options to try: `((MAX_SQL=MIB_LIMIT/4))` or `let MAX_SQL=$MIB_LIMIT/4`

pkg/resource/statefulset.go
```diff
@@ -72,7 +73,7 @@ func (b StatefulSetBuilder) Build(obj client.Object) error {
 		UpdateStrategy: appsv1.StatefulSetUpdateStrategy{
 			RollingUpdate: &appsv1.RollingUpdateStatefulSetStrategy{},
 		},
-		PodManagementPolicy: appsv1.OrderedReadyPodManagement,
+		PodManagementPolicy: appsv1.ParallelPodManagement,
```
chrisseto, line 76: ❤️

apis/v1alpha1/cluster_defaults.go
chrisseto: As resources are just being set via the CRD, couldn't we safely do the math in the operator? Otherwise we'd have to wrap everything in `/bin/sh` to get the envvars to evaluate.

chrisseto (dismissed):
LGTM sans the memory limit issues

apis/v1alpha1/cluster_defaults.go
keith-mcclellan: What would we do in the case where a customer didn't set a resource request or limit?

chrislovecnm: `((MAX_SQL=MIB_LIMIT/4))` won't work.  I am not a shell script expert ... LOL.  Maybe we run the match before we run the command and set it there?

keith-mcclellan: that's worth a try @chrislovecnm

chrisseto: > What would we do in the case where a customer didn't set a resource request or limit?

Either default back to `25%` or not set it at all? It might be better to not set it and then allow users to set the values manually with extra flags?

pkg/kube/helpers.go
```diff
@@ -282,3 +284,60 @@ func GetPodConditionFromList(conditions []corev1.PodCondition, conditionType cor
 	}
 	return -1, nil
 }
+
+func CreateDatabaseServiceAccount(ctx context.Context, clientset *kubernetes.Clientset, namespace string) error {
```
keith-mcclellan, line 288: Can we get a comment here @chrislovecnm please?

keith-mcclellan (changes requested):
Looks like you missed one comment on a function, but otherwise LGTM

pkg/testutil/env/sandbox.go
```diff
@@ -153,19 +155,71 @@ func createNamespace(s Sandbox) error {
...
+	if _, err := s.env.Clientset.RbacV1().ClusterRoles().Create(
+		context.TODO(),
+		&rbac.ClusterRole{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "cockroach-database-role",
+			},
+			Rules: []rbac.PolicyRule{
+				{
+					Verbs:         []string{"use"},
+					APIGroups:     []string{"security.openshift.io"},
```
udnay, line 214: Naive question, what does this do when running on GKE? Is it a `no-op`?

udnay (approved):
Other than question LGTM

pkg/testutil/env/sandbox.go
chrislovecnm: All of these e2e changes I need to back out and put into another PR.  I have the e2e passing on openshift for one of the tests, so it is passing and I will just do a different PR with the sts changes.

chrislovecnm:
Closing for #511

PR #481, opened 2021-05-04
add health checker probe

alinadonisa:
### Current implementation for all rolling restarts (regardless of cluster size):

This will be triggered by a rolling update  on the statefulset, which implies that we are updating one pod at a time.

1. We added `cockroach drain node` as a pre-stop hook on the container initiating a graceful drain.
2. We wait on the k8s readiness (which means `/health?ready=1`) of ALL pods in the the cluster.
3. We check  `_status/vars` on all cockroachdb pods looking for pairs like
`ranges_underreplicated{store="1"} 0` and wait if any are non-zero until all are 0. We can recheck every 10 seconds. 

4. We wait 22 seconds and we  check again `_status/vars` on all cockroachdb pods looking for pairs like
`ranges_underreplicated{store="1"} 0`.
5. Once this checks complete, we update another pod. 
6. Integration of the implementation on rolling restarts and upgrades.

### Work TO DO:

1. I have used exec on the pod to get the metric, maybe I should connect directly to the endpoint using the pod name. Things that that should be considered:
-  `_status/vars` it is the endpoint that Prometheus will scrape, so I cannot exec promql queries. I will need to load all the metrics and parse for the metric `ranges_underreplicated{store="1"} 0` probably. The exec approach that I used allows me to grep on the output and get only the value I need, without extra checks.
2. Tests and improvements.

### DIFF
--- a/e2e/e2e_test.go
+++ b/e2e/e2e_test.go
@@ -107,7 +107,7 @@ func TestCreatesSecureCluster(t *testing.T) {
 	sb.StartManager(t, controller.InitClusterReconcilerWithLogger(testLog))
 
 	builder := testutil.NewBuilder("crdb").WithNodeCount(3).WithTLS().
-		WithImage("cockroachdb/cockroach:v20.2.5").
+		WithImage("cockroachdb/cockroach:v20.2.10").
 		WithPVDataStore("1Gi", "standard" /* default storage class in KIND */)
 
 	create := Step{
@@ -167,7 +167,7 @@ func TestUpgradesMinorVersion(t *testing.T) {
 				current.Spec.Image.Name = "cockroachdb/cockroach:v20.2.9"
 				require.NoError(t, sb.Update(current))
 
-				RequireClusterToBeReadyEventuallyTimeout(t, sb, builder, 600*time.Second)
+				RequireClusterToBeReadyEventuallyTimeout(t, sb, builder, 500*time.Second)
 				requireDbContainersToUseImage(t, sb, current)
 				t.Log("Done with upgrade")
 			},
@@ -177,7 +177,7 @@ func TestUpgradesMinorVersion(t *testing.T) {
 	steps.Run(t)
 }
 
-func TestUpgradesMajorVersion19to20(t *testing.T) {
+func TestUpgradesMajorVersion20to21(t *testing.T) {
 
 	// We are doing a major version upgrade here
 	// 19 to 20
@@ -197,7 +197,7 @@ func TestUpgradesMajorVersion19to20(t *testing.T) {
 	sb.StartManager(t, controller.InitClusterReconcilerWithLogger(testLog))
 
 	builder := testutil.NewBuilder("crdb").WithNodeCount(3).WithTLS().
-		WithImage("cockroachdb/cockroach:v19.2.6").
+		WithImage("cockroachdb/cockroach:v20.2.9").
 		WithPVDataStore("1Gi", "standard" /* default storage class in KIND */)
 
 	steps := Steps{
@@ -215,7 +215,7 @@ func TestUpgradesMajorVersion19to20(t *testing.T) {
 				current := builder.Cr()
 				require.NoError(t, sb.Get(current))
 
-				current.Spec.Image.Name = "cockroachdb/cockroach:v20.1.1"
+				current.Spec.Image.Name = "cockroachdb/cockroach:v21.1.0"
 				require.NoError(t, sb.Update(current))
 
 				RequireClusterToBeReadyEventuallyTimeout(t, sb, builder, 500*time.Second)
@@ -229,8 +229,6 @@ func TestUpgradesMajorVersion19to20(t *testing.T) {
 }
 
 func TestUpgradesMajorVersion20_1To20_2(t *testing.T) {
-
-	// Major version upgrade 20_1 to 20_2
 
 	if parallel {
 		t.Parallel()
@@ -247,7 +245,7 @@ func TestUpgradesMajorVersion20_1To20_2(t *testing.T) {
 	sb.StartManager(t, controller.InitClusterReconcilerWithLogger(testLog))
 
 	builder := testutil.NewBuilder("crdb").WithNodeCount(3).WithTLS().
-		WithImage("cockroachdb/cockroach:v20.1.12").
+		WithImage("cockroachdb/cockroach:v20.1.16").
 		WithPVDataStore("1Gi", "standard" /* default storage class in KIND */)
 
 	steps := Steps{
@@ -265,9 +263,10 @@ func TestUpgradesMajorVersion20_1To20_2(t *testing.T) {
 				current := builder.Cr()
 				require.NoError(t, sb.Get(current))
 
-				current.Spec.Image.Name = "cockroachdb/cockroach:v20.2.5"
+				current.Spec.Image.Name = "cockroachdb/cockroach:v20.2.10"
 				require.NoError(t, sb.Update(current))
-
+				// we wait 10 min because we will be waiting 3 min for each pod because
+				// v20.1.16 does not have curl installed
 				RequireClusterToBeReadyEventuallyTimeout(t, sb, builder, 500*time.Second)
 				requireDbContainersToUseImage(t, sb, current)
 				t.Log("Done with major upgrade")
@@ -340,7 +339,7 @@ func TestPVCResize(t *testing.T) {
 	sb := testenv.NewDiffingSandbox(t, env)
 	sb.StartManager(t, controller.InitClusterReconcilerWithLogger(testLog))
 	builder := testutil.NewBuilder("crdb").WithNodeCount(3).WithTLS().
-		WithImage("cockroachdb/cockroach:v20.2.5").
+		WithImage("cockroachdb/cockroach:v20.2.10").
 		WithPVDataStore("1Gi", "standard" /* default storage class in KIND */)
 	steps := Steps{
 		{
--- a/pkg/actor/cluster_restart.go
+++ b/pkg/actor/cluster_restart.go
@@ -25,6 +25,7 @@ import (
 	api "github.com/cockroachdb/cockroach-operator/apis/v1alpha1"
 	"github.com/cockroachdb/cockroach-operator/pkg/condition"
 	"github.com/cockroachdb/cockroach-operator/pkg/features"
+	"github.com/cockroachdb/cockroach-operator/pkg/healthchecker"
 	"github.com/cockroachdb/cockroach-operator/pkg/resource"
 	"github.com/cockroachdb/cockroach-operator/pkg/scale"
 	"github.com/cockroachdb/cockroach-operator/pkg/utilfeature"
@@ -90,6 +91,7 @@ func (r *clusterRestart) Act(ctx context.Context, cluster *resource.Cluster) err
 	if err != nil {
 		return errors.Wrapf(err, "failed to create kubernetes clientset")
 	}
+
 	statefulSet := &appsv1.StatefulSet{}
 	if err := r.client.Get(ctx, key, statefulSet); err != nil {
 		return errors.Wrap(err, "failed to fetch statefulset")
@@ -99,16 +101,15 @@ func (r *clusterRestart) Act(ctx context.Context, cluster *resource.Cluster) err
 	if statefulSetIsUpdating(statefulSet) {
 		return NotReadyErr{Err: errors.New("restart statefulset is updating, waiting for the update to finish")}
 	}
-
 	status := &statefulSet.Status
 	if status.CurrentReplicas == 0 || status.CurrentReplicas < status.Replicas {
 		log.Info("restart statefulset does not have all replicas up")
 		return NotReadyErr{Err: errors.New("restart cluster statefulset does not have all replicas up")}
 	}
-
+	healthChecker := healthchecker.NewHealthChecker(cluster, clientset, r.scheme, r.config)
 	if strings.EqualFold(restartType, api.ClusterRestartType(api.RollingRestart).String()) {
 		log.V(DEBUGLEVEL).Info("initiating rolling restart action")
-		if err := r.rollingSts(ctx, statefulSet.DeepCopy(), clientset, r.log); err != nil {
+		if err := r.rollingSts(ctx, statefulSet.DeepCopy(), clientset, r.log, healthChecker); err != nil {
 			return errors.Wrapf(err, "error restarting statefulset %s.%s", cluster.Namespace(), cluster.StatefulSetName())
 		}
 		log.V(DEBUGLEVEL).Info("completed rolling cluster restart")
@@ -118,8 +119,9 @@ func (r *clusterRestart) Act(ctx context.Context, cluster *resource.Cluster) err
 		}
 		//sleep 1 minute to make sure the crdb is up and running
 		log.V(DEBUGLEVEL).Info("sleeping", "duration", sleepDuration.String(), "label", "after full cluster restart")
-		time.Sleep(sleepDuration)
-
+		if err := healthChecker.Probe(ctx, log, fmt.Sprintf("waiting after restart for cluster %s", cluster.Name()), 0); err != nil {
+			return err
+		}
 		log.V(DEBUGLEVEL).Info("completed full cluster restart")
 	} else {
 		err := ValidationError{Err: errors.New("invalid annotation value, please use Rolling or FullCluster values")}
@@ -147,7 +149,10 @@ func (r *clusterRestart) Act(ctx context.Context, cluster *resource.Cluster) err
 }
 
 // rollingSts performs a rolling update on the cluster.
-func (r *clusterRestart) rollingSts(ctx context.Context, sts *appsv1.StatefulSet, clientset *kubernetes.Clientset, l logr.Logger) error {
+func (r *clusterRestart) rollingSts(ctx context.Context, sts *appsv1.StatefulSet,
+	clientset *kubernetes.Clientset,
+	l logr.Logger,
+	healthChecker healthchecker.HealthChecker) error {
 	timeNow := metav1.Now()
 	// When a StatefulSet's partition number is set to `n`, only StatefulSet pods
 	// numbered greater or equal to `n` will be updated. The rest will remain untouched.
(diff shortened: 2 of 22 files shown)


keith-mcclellan:
1. We update one pod at a time.
2. `cockroach drain node` is sent first as a pre-stop hook initiating a graceful drain, after 1m, SIGKILL is sent.
3. We wait on the k8s readiness (which means /health?ready=1) of ALL pods in the the cluster.
4. we check  _status/vars on all nodes looking for pairs like
ranges_underreplicated{store="1"} 0 and wait if any are non-zero until all are 0. We can recheck every 10 seconds. 
5. once the above is true we wait x seconds and check a 2nd time.  10 seconds is the period under which we guarantee updated prometheus stats so that should be the minimum. Currently, we wait 1 minute but without an additional health check, in CC the default is 3 minutes.  We want the user to be able to set the safety window and will document the tradeoffs on a longer window.
6. Once this checks complete, we update another pod.

@udnay thoughts? @tbg would love your input as well

udnay:
This makes sense to me. I do have one question why are we waiting a second time and then rechecking before continuing? Do we expect the replicated ranges to potentially blip after they come back as ok?

keith-mcclellan:
The periodicity of the prometheus refresh is 10 seconds, so if a replica rejoined their raft group and then got ejected for some reason we won't roll to the next pod.  That shouldn't happen very often but theoretically it can in cases where you have clock skew issues or if it got too stale and we started rehydrating a replacement.

chrislovecnm:
@keith-mcclellan how many times do we test prometheus till we fail?

udnay:
I was thinking we check until we hit our current timeout. So waiting would be between 1m and our timeout. It would be nice if the timeout was eventually configurable

keith-mcclellan:
I agree - if it hasn't passed at least once by the timeout we should roll back.  I think we need to set the default there to 3m

alinadonisa:
> Proposed state for all rolling restarts (regardless of cluster size):
> 
> 1. We update one pod at a time.
> 2. `cockroach drain node` is sent first as a pre-stop hook initiating a graceful drain, after 1m, SIGKILL is sent.
> 3. We wait on the k8s readiness (which means `/health?ready=1`) of ALL pods in the the cluster.
> 4. we check  `_status/vars` on all cockroachdb pods looking for pairs like
>    `ranges_underreplicated{store="1"} 0` and wait if any are non-zero until all are 0. We can recheck every 10 seconds.
> 5. once the above is true we wait x seconds and check a 2nd time.  10 seconds is the period under which we guarantee updated prometheus stats so that should be the minimum. Currently, we wait 1 minute but without an additional health check, in CC the default is 3 minutes.  We want the user to be able to set the safety window and will document the tradeoffs on a longer window.
> 6. Once this checks complete, we update another pod.
> 
> @udnay thoughts? @tbg would love your input as well

@keith-mcclellan  after I exec on any  pod of the cluster  and try to get the ranges_underreplicated metric I noticed that the store label is actually the id of the pod. Just to make sure. Is this correct? 
```bash
curl -Lk localhost:8080/_status/vars --silent | grep -i  ranges_underreplicated
# HELP ranges_underreplicated Number of ranges with fewer live replicas than the replication target
# TYPE ranges_underreplicated gauge
ranges_underreplicated{store="3"} 0
```

pkg/healthchecker/healthchecker.go
```diff
@@ -18,61 +18,152 @@ package healthchecker
...
+		cluster:   cluster,
+		config:    config,
+	}
 }
 
-func (s *HealthCheckerImpl) Probe(ctx context.Context, l logr.Logger, logSuffix string) error {
-	l.V(int(zapcore.DebugLevel)).Info("Health check probe", "label", logSuffix)
-	return s.waitUntilAllReplicasAreRunning(ctx, l, logSuffix)
-}
+func (s *HealthCheckerImpl) Probe(ctx context.Context, l logr.Logger, logSuffix string, nodeID int) error {
```
alinadonisa, line 66: @keith-mcclellan , @udnay , @chrislovecnm please check the logic from the Probe method. Take in consideration that this runs after the rolling update is triggered. The prestop hook already runs when thisProbe method is called so we wait 1 min for the terminationGracePeriodSeconds to end and we start our logic. I commented the sendKillSignal function because we just restarted.

pkg/healthchecker/healthchecker.go
```diff
@@ -0,0 +1,169 @@
...
+	//we sleep 1 minunte to ensure the pre-stop hook finished
+	time.Sleep(1 * time.Minute)
+	//SKIP KILL because we just restarted????
+	// err := sendKillSignal(ctx, l, logSuffix, podname, nodeID)
+	// if err != nil {
+	// 	return err
+	// }
+
+	//isready check for all pods from sts
+	err := kube.WaitUntilAllStsPodsAreReady(ctx, s.clientset, l, stsname, stsnamespace, 30*time.Minute, 10*time.Minute)
```
udnay, line 81: 30 minutes seems excessive. I think after 3m we either fail or move one

pkg/healthchecker/healthchecker.go
```diff
@@ -0,0 +1,169 @@
...
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+//sendKillSignal it is not necessary in our scenario... we just restarted the node and run drain as pre-stop hook
+func sendKillSignal(ctx context.Context, l logr.Logger, logSuffix, podname string, nodeID int) error {
```
udnay, line 103: If we don't need this anymore lets delete it

pkg/healthchecker/healthchecker.go
udnay: So the order here is
1. restart is triggered
2. prestop hook runs calling `cockroach drain node`
3. probe gets called immediately so we sleep 1min?

pkg/healthchecker/healthchecker.go
```diff
@@ -0,0 +1,169 @@
...
+//checkUnderReplicatedMetric uses exec on the pod for now
+func (s *HealthCheckerImpl) checkUnderReplicatedMetric(ctx context.Context, l logr.Logger, logSuffix, podname string, partition int32) error {
+	l.V(int(zapcore.DebugLevel)).Info("checkUnderReplicatedMetric", "label", logSuffix, "podname", podname, "partition", partition)
+	port := strconv.FormatInt(int64(*s.cluster.Spec().HTTPPort), 10)
+	underrepmetric := fmt.Sprintf(underreplicatedmetric, partition)
+	cmd := []string{
+		fmt.Sprintf(cmdunderreplicted, port, underrepmetric),
+	}
+	l.V(int(zapcore.DebugLevel)).Info("get ranges_underreplicated metric", "node", podname)
+	output, stderr, err := kube.ExecInPod(s.scheme, s.config, s.cluster.Namespace(),
```
udnay, line 139: We should be hitting the prometheus endpoint directly.

pkg/actor/cluster_restart.go
```diff
@@ -239,7 +189,7 @@ func (r *clusterRestart) rollingSts(ctx context.Context, sts *appsv1.StatefulSet
 		}
 
 		// wait for all replicas to be up
-		healthChecker.Probe(ctx, l, "between restarting pods")
+		healthChecker.Probe(ctx, l, "between restarting pods", int(partition))
```
udnay, line 192: Because we wait for the pod to be ready `WaitUntilStatefulSetIsReadyToServe` do we actually need to sleep for a minute? If we do, I'd rather wait here and keep the probe better scoped to what it's trying to accomplish

pkg/healthchecker/healthchecker.go
udnay: Also what's the difference between this method and `WaitUntilStatefulSetIsReadyToServe`

pkg/healthchecker/healthchecker.go
```diff
@@ -18,61 +18,152 @@ package healthchecker
...
-func (s *HealthCheckerImpl) waitUntilAllReplicasAreRunning(ctx context.Context, l logr.Logger, logSuffix string) error {
+
+func (s *HealthCheckerImpl) waitUntilUnderReplicatedMetricIsZero(ctx context.Context, l logr.Logger, logSuffix, podname string, partition int32) error {
 	f := func() error {
-		return s.checkAllReplicasAreRunning(ctx, l, logSuffix)
+		return s.checkUnderReplicatedMetric(ctx, l, logSuffix, podname, partition)
 	}
 	b := backoff.NewExponentialBackOff()
-	b.MaxElapsedTime = 30 * time.Minute
+	b.MaxElapsedTime = 3 * time.Minute
```
udnay, line 123: For this we just want to do a poll every 10 seconds, maybe with a bit of jitter but should be fine as we hit each node independently. If we don't get 0 for all nodes within 3 minutes then we should stop.

pkg/actor/cluster_restart.go
alinadonisa: Yes... we need this, because we know that the terminationGracePeriodSeconds is 60 seconds, and has no point to make requests

pkg/healthchecker/healthchecker.go
alinadonisa: this works as expected

pkg/healthchecker/healthchecker.go
alinadonisa: totally agree :)

pkg/healthchecker/healthchecker.go
alinadonisa: yes

pkg/healthchecker/healthchecker.go
alinadonisa: I added kube.WaitUntilAllStsPodsAreReady in kube package that is common so we use it in all the packages.  Actually WaitUntilStatefulSetIsReadyToServe  will be replace with kube.WaitUntilAllStsPodsAreReady.

chrislovecnm:
Here are the wait loops

```
func check() bool, customError {
  loop for three minutes {
   test, err := checkAllPodsPromethus()
   if err return false, err
   if test {
        sleep(12)
        test, err := checkAllpods()
        if err return false, err
        if test {
          return true, nil
        } else {
         return false, customError 
       }
    } else {
       return false, customError 
    }
  }
  // loop ended no happy path
  return false, customError
}

func checkallpods() bool, error {
   sleep(10)
   return loop through all pods and check prometheus
}
```

keith-mcclellan:
> if test {
>         sleep(12)
>         test, err := checkAllpods()
>         if err return false, err
>         if test {
>           return true, nil
>         } else {
>          return false, customError

LGTM

chrislovecnm:
```bash
curl -ks https://${POD_NAME}.${DATABASE_NAME}:8080/_status/vars | grep 'ranges_underreplicated{'
```

Is the example

udnay:
pseudo code LGTM

pkg/actor/cluster_restart.go
```diff
@@ -185,9 +190,10 @@ func (r *clusterRestart) rollingSts(ctx context.Context, sts *appsv1.StatefulSet
 			return errors.Wrapf(err, "error rolling update stategy on pod %d", int(partition))
 		}
 
-		// wait 1 minute between updates
-		l.V(int(zapcore.DebugLevel)).Info("sleeping", "duration", sleepDuration.String(), "label", "between restarting pods")
-		time.Sleep(sleepDuration)
+		// wait for all replicas to be up
+		if err := healthChecker.Probe(ctx, l, "between restarting pods", int(partition)); err != nil {
```
udnay, line 194: We should maybe add an issue to add the `healthchecker` in the full restart as well. There we only wait on the `scale.WaitUntilStatefulSetIsReadyToServe` to finish. The DB is unavailable for that but it might be valuable to ensure everything is good before finishing the op. wdyt @keith-mcclellan?

@alinadonisa the above does not need to be a part of this PR though

keith-mcclellan: fullRestart actually restarts all the nodes at once - it's intended largely to handle cases where we know we need to take an outage (like changing CA's for the node cert) to fix something. That action will always cause database unavailability so I'm not sure what adding this to that restart would do @udnay ?  The only utility I could see is it could be used to block other admin actions by the CR until the database is completely healthy, but I'd think we would fix that in the refactor using an admission webhook rather than continuing to enhance this restart actor.

pkg/resource/testdata/TestStatefulSetBuilder/insecure_statefulset_with_resources.golden
```diff
@@ -73,6 +73,11 @@ spec:
             scheme: HTTP
           initialDelaySeconds: 10
           periodSeconds: 5
+        lifecycle:
+          preStop:
```
udnay, line 78: Naive question, is the 1m grace period the default or set somewhere else, should we be setting it explicitly here?

udnay (approved):
Couple of nits but LGTM

pkg/actor/resize_pvc.go
```diff
@@ -260,25 +259,3 @@ func (rp *resizePVC) findAndResizePVC(ctx context.Context, sts *appsv1.StatefulS
 	log.Info("found and resized all PVCs")
 	return nil
 }
-
-// rollSts performs a rolling update on the cluster.
-func (rp *resizePVC) rollSts(ctx context.Context, cluster *resource.Cluster, clientset *kubernetes.Clientset) error {
```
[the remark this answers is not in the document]
chrislovecnm: never got called. Removing dead code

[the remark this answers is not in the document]
keith-mcclellan: looking at lines 139-164, I don't think the restart actually gets called - theres stub code and a TODO.  Can we please manually test this on GKE @alinadonisa and make sure we are triggering a restart ?  cc: @chrislovecnm

pkg/resource/testdata/TestStatefulSetBuilder/insecure_statefulset_with_resources.golden
keith-mcclellan: we need to add `terminationGracePeriodSeconds` and set it to 60 seconds here good catch @udnay

pkg/resource/testdata/TestStatefulSetBuilder/insecure_statefulset_cli_args.golden
```diff
@@ -74,6 +74,11 @@ spec:
             scheme: HTTP
           initialDelaySeconds: 10
           periodSeconds: 5
+        lifecycle:
+          preStop:
```
keith-mcclellan, line 78: we need to add terminationGracePeriodSeconds and set it to 60 seconds here

pkg/resource/testdata/TestStatefulSetBuilder/default_secure.golden
```diff
@@ -74,6 +74,11 @@ spec:
             scheme: HTTPS
           initialDelaySeconds: 10
           periodSeconds: 5
+        lifecycle:
+          preStop:
```
keith-mcclellan, line 78: we need to add terminationGracePeriodSeconds and set it to 60 seconds here

pkg/resource/testdata/TestStatefulSetBuilder/default_insecure_statefulset.golden
```diff
@@ -74,6 +74,11 @@ spec:
             scheme: HTTP
           initialDelaySeconds: 10
           periodSeconds: 5
+        lifecycle:
+          preStop:
```
keith-mcclellan, line 78: we need to add terminationGracePeriodSeconds and set it to 60 seconds here

keith-mcclellan (changes requested):
We need to manually test resizing a PVC with this change (since we can't on kind), and the preStop hooks all need to add terminationGracePeriodSeconds and set it to 60 seconds

pkg/actor/resize_pvc.go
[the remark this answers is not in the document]
chrislovecnm: @keith-mcclellan the code would not compile if we broke this

@rail can we run gke nightly against this PR easily?

pkg/healthchecker/healthchecker.go
```diff
@@ -0,0 +1,181 @@
...
+	port := strconv.FormatInt(int64(*s.cluster.Spec().HTTPPort), 10)
+	store := partition + 1
+	underrepmetric := fmt.Sprintf(underreplicatedmetric, int(store))
+	cmd := []string{
+		"/bin/bash",
+		"-c",
+		fmt.Sprintf(cmdunderreplicted, podname, stsname, stsnamespace, port),
+	}
+	l.V(int(zapcore.DebugLevel)).Info("get ranges_underreplicated metric", "node", podname, "underrepmetric", underrepmetric, "cmd", cmd)
+	output, stderr, err := kube.ExecInPod(s.scheme, s.config, s.cluster.Namespace(),
```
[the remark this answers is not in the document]
chrislovecnm: Actually using the exec makes e2e easier. We cannot make an http call to a pod external to the cluster. Thoughts?

pkg/resource/testdata/TestStatefulSetBuilder/default_secure.golden
alinadonisa: @chrislovecnm  it is already in there, but at the pod spec level not container level ... see https://github.com/cockroachdb/cockroach-operator/blob/master/pkg/resource/statefulset.go#L198

keith-mcclellan (approved):
LGTM pending e2e success

udnay (dismissed):
:shipit:

pkg/actor/resize_pvc.go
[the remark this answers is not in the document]
rail: Let me see...

e2e/e2e_test.go
```diff
@@ -218,7 +218,7 @@ func TestUpgradesMajorVersion20to21(t *testing.T) {
 				current.Spec.Image.Name = "cockroachdb/cockroach:v21.1.0"
 				require.NoError(t, sb.Update(current))
 
-				RequireClusterToBeReadyEventuallyTimeout(t, sb, builder, 500*time.Second)
+				RequireClusterToBeReadyEventuallyTimeout(t, sb, builder, 600*time.Second)
```
chrislovecnm: We need these backed out

alinadonisa: @chris it is already rolled back. See https://github.com/cockroachdb/cockroach-operator/blob/5b90ae2b235d3883d4532ada51a6c7aa6612e415/e2e/e2e_test.go#L221

pkg/actor/resize_pvc.go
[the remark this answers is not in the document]
rail: It's running now: https://teamcity.cockroachdb.com/viewLog.html?buildId=3011994&buildTypeId=Cockroach_CockroachOperator_Nightlies_GkeNightly

PR #488, opened 2021-05-06
Rolling restart e2e test

udnay:
Testing the rolling restart functionality with e2e test. Validates that
the pods at the end of the test all have a newer timestamp than when the
cluster was initialized.

### DIFF
--- a/e2e/e2e_test.go
+++ b/e2e/e2e_test.go
@@ -20,12 +20,14 @@ import (
 	"flag"
 	"os"
 	"path/filepath"
+	"strings"
 	"testing"
 	"time"
 
 	api "github.com/cockroachdb/cockroach-operator/apis/v1alpha1"
 	"github.com/cockroachdb/cockroach-operator/pkg/actor"
 	"github.com/cockroachdb/cockroach-operator/pkg/controller"
+	"github.com/cockroachdb/cockroach-operator/pkg/labels"
 	"github.com/cockroachdb/cockroach-operator/pkg/testutil"
 	testenv "github.com/cockroachdb/cockroach-operator/pkg/testutil/env"
 	"github.com/cockroachdb/cockroach-operator/pkg/testutil/paths"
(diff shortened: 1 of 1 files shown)


e2e/e2e_test.go
```diff
@@ -368,3 +372,100 @@ func TestPVCResize(t *testing.T) {
...
+
+	testLog := zapr.NewLogger(zaptest.NewLogger(t))
+
+	actor.Log = testLog
+
+	sb := testenv.NewDiffingSandbox(t, env)
+	sb.StartManager(t, controller.InitClusterReconcilerWithLogger(testLog))
+
+	builder := testutil.NewBuilder("crdb").WithNodeCount(3).WithTLS().
+		WithImage("cockroachdb/cockroach:v20.2.5").
```
keith-mcclellan, line 394: We need an enhancement to the testing framework that pulls this from a central place we can keep up to date.  There are relatively few tests where we should be pinning to specific releases and those are largely around upgrades

e2e/e2e_test.go
```diff
@@ -368,3 +372,100 @@ func TestPVCResize(t *testing.T) {
...
+					} else if rt == "" { // The annotation was deleted
+						return true, nil
+					}
+
+					return false, nil
+				})
+				require.NoError(t, err)
+
+				// Make sure the cluster is in a good state after the restart is finished
+				RequireClusterToBeReadyEventuallyTimeout(t, sb, builder, 500*time.Second)
```
keith-mcclellan, line 445: I think your timeout here will be too short if we merge the safer restart code Alina is working on.  This is a great reason why we need to do a better job parallelizing these test cases.

e2e/e2e_test.go
```diff
@@ -368,3 +372,100 @@ func TestPVCResize(t *testing.T) {
...
+						continue // Skip the validate version pod
+					}
+					podCreationTimes[pod.Name] = pod.CreationTimestamp
+				}
+
+				current.Annotations["crdb.io/restarttype"] = "rolling"
+				require.NoError(t, sb.Update(current))
+
+				// Need to poll for getting rid of the restart annotation
+				err = wait.Poll(10*time.Second, 500*time.Second, func() (done bool, err error) {
```
keith-mcclellan, line 428: I think this is likely to fail once we add the enhanced restart logic because of a timeout issue

keith-mcclellan (changes requested):
There could be other items but I noticed a few spots where I'd expect this to fall over, particularly after we merge Alina's enhanced health check PR

udnay:
Going to close this as we test restart a lot implicitly already and our e2e tests need some rethinking

PR #493, opened 2021-05-13
use cockroachdb namespace

taroface:
Change the Operator manifest to use `cockroachdb` instead of `default` namespace.

### DIFF
--- a/manifests/operator.yaml
+++ b/manifests/operator.yaml
@@ -29,7 +29,7 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: cockroach-database-sa
-  namespace: default
+  namespace: cockroachdb
   annotations:
   labels:
     app: cockroach-operator
@@ -46,7 +46,7 @@ roleRef:
 subjects:
   - kind: ServiceAccount
     name: cockroach-database-sa
-    namespace: default
+    namespace: cockroachdb
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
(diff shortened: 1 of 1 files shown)


chrislovecnm:
@taroface this might break our GKE e2e testing.  I am uncertain if the `make k8s/apply` will create the namespace.  Can you please test it?

taroface:
@chrislovecnm This manifest works for me. I ran `kubectl create namespace cockroachdb` and then applied the `operator.yaml` using `namespace: cockroachdb` throughout.

`kubectl get pods --all-namespaces` shows:

```
cockroachdb   cockroach-operator-64fb9fd497-tjrc2                        1/1     Running   0          2m27s
cockroachdb   cockroachdb-0                                              1/1     Running   0          99s
cockroachdb   cockroachdb-1                                              1/1     Running   0          76s
cockroachdb   cockroachdb-2                                              1/1     Running   0          51s
```

chrislovecnm:
The e2e test does not run that kubectl command to create the ns, can you run the apply without the create ns?

taroface:
OK, it does error out with `namespaces "cockroachdb" not found`. @keith-mcclellan can you weigh in here?

keith-mcclellan:
@taroface I guess we need to stick with default for the launch on GKE, but let's backlog an issue to make this change to the build so we can have the default me a custom namespace.

If you want, we can document manually updating the configs to use a custom namespace - I expect that will be a common request.

taroface:
Closing this issue in favor of #495.

PR #494, opened 2021-05-13
uncomment resources, update image

### LINKED ISSUE #463: Operator example.yaml has incorrect indentation
During the bug bash, we found that the `example.yaml` file had the hardware resource limits at the same level as `pvc`. This should instead be aligned with `dataStore` to avoid causing issues when the lines are uncommented.

taroface:
Fixes #463.

### DIFF
--- a/examples/example.yaml
+++ b/examples/example.yaml
@@ -34,7 +34,7 @@ spec:
       cpu: "2"
       memory: "8Gi"
   tlsEnabled: true
-  # cockroachDBVersion: v20.2.8
+# cockroachDBVersion: v20.2.8
   image:
     name: cockroachdb/cockroach:v20.2.8
   nodes: 3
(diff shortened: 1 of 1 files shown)


chrislovecnm (commented):
You either use one or the other, not both

taroface:
> You either use one or the other, not both

I don't know what this comment means. This PR only uncomments the `resources` object.

I also moved `cockroachDBVersion` to the left because there was confusion about the alignment of commented-out fields. This field is still commented out in favor of `image`.

chrislovecnm:
`cockroachDBVersion` I missed that it was still commented out :)

chrislovecnm (approved):
LGTM

keith-mcclellan (commented):
LGTM

PR #508, opened 2021-05-17
update CRDB version to v21.1.0 in example.yaml

taroface:
Updated CRDB version to v21.1.0.


taroface:
Closing as I think this PR inherited some stale updates.

PR #510, opened 2021-05-18
update logo in OpenShift UI

keith-mcclellan:
We need to update the logo in the OpenShift UI to the latest branding and it's base64 encoded in the CSV.


udnay (approved):
I assume it doesn't need to be svg

keith-mcclellan:
> I assume it doesn't need to be svg

PNG should be fine - see https://docs.openshift.com/container-platform/4.7/operators/operator_sdk/osdk-generating-csvs.html

Issue #417, opened 2021-04-15
need to set cache values discretely not relatively


keith-mcclellan:
In https://github.com/cockroachdb/cockroach-operator/blob/1c11d981aadfaf65699bd5977899c74f1b7801f1/apis/v1alpha1/cluster_defaults.go, we are setting each of the caches to a relative value of 25%, but in many cases the total memory represented to the pod isn't accurate so we need to discretely set these values. see https://github.com/cockroachdb/cockroach/pull/51471/commits/1456ed178865dd435ba35c97c967a3dbf720ca6a as an example for how we recommend that we set these now.

keith-mcclellan:
@taroface caught this during his docs review process

Issue #470, opened 2021-04-29
Statefulsets should use `Parallel` pod management


chrisseto:
The underlying statefulset that is created by the operator is has its pod management policy set to `OrderReady` rather than `Parallel`. We've seen that `OrderReady` can lead to cluster unavailability when all node happen to crash within a short interval.

For example, if a query can cause nodes to OOM (more comment in previous CRDB iterations) it may very well take down nodes one after the other if it's being continuously retried. In such a case, pod ordinal 0 may never mark itself as ready due to the inability to contact other CRDB nodes. When using `OrderReady` the other CRDB nodes won't be rescheduled until pods with a lower ordinal have marked themselves as ready, leading to a deadlock.

chrislovecnm:
Can we close this @johnrk?

chrisseto:
Closed via https://github.com/cockroachdb/cockroach-operator/commit/998fbf7f18c9f4b27ae6de1043f980739956bc5e

PR #511, opened 2021-05-18
StatefulSet Changes - CR Cache, Memory, and sts pod management

chrislovecnm:
1. Dynamically set cache and max sql memory if the user does
not set a CRD value
2. sts is set to Parallel pod management

We are making changes to use `exec` and bash directly in order to
start the Pods and set the cache values dynamically.  We are also
adding environment variables in to the Pods that set the GOPROCSMAX
and also MEMORY_LIMIT_MIB which is used to set CR cachee and
max sql memory values.


udnay:
this fixes #417 and 470 right @keith-mcclellan?

keith-mcclellan (approved):
LGTM

keith-mcclellan:
> this fixes #417 and #470 right @keith-mcclellan?

Yes!

PR #513, opened 2021-05-19
Decommission needs to run first

keith-mcclellan:
Adjusting the order that the actors run to execute Decommission first so that upgrade doesn't interrupt it.

### DIFF
--- a/pkg/actor/actor.go
+++ b/pkg/actor/actor.go
@@ -80,6 +80,8 @@ type Actor interface {
 func NewOperatorActions(scheme *runtime.Scheme, cl client.Client, config *rest.Config) []Actor {
 
 	var update Actor
+        // move decommission to top to avoid conflict with other actors
+	decommission := newDecommission(scheme, cl, config)
 
 	// entry point for new PartitionUpdate upgrades
 	// this feature is controlled by a featuregate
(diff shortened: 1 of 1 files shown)


pkg/actor/actor.go
```diff
@@ -89,7 +91,7 @@ func NewOperatorActions(scheme *runtime.Scheme, cl client.Client, config *rest.C
 		update = newUpgrade(scheme, cl, config)
 	}
 
-	decommission := newDecommission(scheme, cl, config)
```
chrislovecnm, line 92: This line does not need to change. It is the order in the slice that matters

pkg/actor/actor.go
```diff
@@ -110,10 +112,11 @@ func NewOperatorActions(scheme *runtime.Scheme, cl client.Client, config *rest.C
 	// Actors that controlled by featuregates
 	// have the featuregate check above or in there handles
 	// func.
+	// decommission needs to be first, it is not dependant on versionchecker
 	return []Actor{
+		decommission,
```
chrislovecnm, line 117: This is where it needed to change 😀 - this is correct

udnay (changes requested):
Change looks ok to me, but can you fix up the commit history so we keep master clean?

The other thing, should the `Handle` functions be more restrictive to make sure this type of thing doesn't happen again?

keith-mcclellan:
> Change looks ok to me, but can you fix up the commit history so we keep master clean?
> 

Yeah I got into weird rebase hell, I'll get the commit history cleaned up tomorrow.

keith-mcclellan:
> The other thing, should the `Handle` functions be more restrictive to make sure this type of thing doesn't happen again?
@chrislovecnm can you chime in here? I would assume the webhooks from the refactor would stop this from happening in the future

udnay:
Are there times when we want multiple actors to fire for a single event? I guess restart?

Issue #473, opened 2021-04-29
Major upgrades after scaling fail.


alyshanjahani-crl:
I created a 5 node cluster (on v20.1.1) w/ the operator. Then Scaled to 3.
Then change the image to v20.2.1.
Let operator update the image of the crdb pods in the cluster.
Then `set cluster setting cluster.preserve_downgrade_option=''` to finalize the upgrade.

The finalization was unsuccessful, relevant logs from the crdb pods:
```
I210429 15:17:30.338514 1530 server/server_update.go:52 ⋮ [n1] failed attempt to upgrade cluster version, error: node 4 not running (‹NODE_STATUS_DEAD›), cannot determine version
```

```
**I210429 15:24:01.902099 71244 rpc/nodedialer/nodedialer.go:160 ⋮ [n2] unable to connect to n5: failed to connect to n5 at ‹cockroachdb-4.cockroachdb.default:26258›: ‹initial connection heartbeat failed›: ‹rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing dial tcp: lookup cockroachdb-4.cockroachdb.default on 10.3.240.10:53: no such host"›
I210429 15:24:03.702154 71284 rpc/nodedialer/nodedialer.go:160 ⋮ [n2] unable to connect to n4: failed to connect to n4 at ‹cockroachdb-3.cockroachdb.default:26258›: ‹initial connection heartbeat failed›: ‹rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing dial tcp: lookup cockroachdb-3.cockroachdb.default on 10.3.240.10:53: no such host"**
```

johnrk:
@alinadonisa will investigate this issue.

alinadonisa:
@alyshanjahani-crl  when you scaled to 3 nodes, node 4 and 5 were deleted, but somehow the cockroach nodes that remained still try to connect to the  deleted  ones. The problem is on decommission, the node 4 and 5 did not drained correctly. @chrisseto  do you have a fix for this?

keith-mcclellan:
@alinadonisa how is the operator getting the list of nodes that need to be upgraded?

alinadonisa:
@keith-mcclellan because we are using RollingUpdate strategy of the Statefulset, we are making the checks for each pod in the sts iterating through sts.Spec.Replicas.
 We get each po using:
```golang
podName := fmt.Sprintf("%s-%d", stsName, podNumber)

[...]

PR #514, opened 2021-05-20
Update actor order to run decommission first

keith-mcclellan:
Decommission needs to run before other actors

This should resolve the issues we are seeing on https://github.com/cockroachdb/cockroach-operator/issues/473


keith-mcclellan:
@udnay @chrislovecnm my git foo wasn't good enough to fix the commit history so I just did a new PR

udnay (approved):
LGTM

PR #515, opened 2021-05-20
Removing default values

chrislovecnm:
Previously we set defaults for cache and max sql memory. We now
set those values based on Pod limits. The cluster_defaults.go
code was removed in a different PR, but the PR that was merged
did not have those changes. These changes are coverred in a
unit test, but the test data was incorrectly updated.


keith-mcclellan (approved):
LGTM

PR #521, opened 2021-05-24
Creating RBAC objects for Openshift E2E

chrislovecnm:
This PR creates the RBAC bindings that are required to run e2e testing
with openshift. We are creating the role binding and role that
allows the operator database service account to a Pod as any UID.

When the e2e test starts the role is created, and when the test
is over the role is destroyed.  This code stop e2e testing
to occur in parallel. But we have other problems with parallel testing
as well.


keith-mcclellan (dismissed):
LGTM, but sandbox.go needs more comments so please open another PR with that.

chrislovecnm:
> LGTM, but sandbox.go needs more comments so please open another PR with that.

Doesn't that require an issue???

keith-mcclellan (approved):
LGTM

PR #522, opened 2021-05-25
Removing logging library

chrislovecnm:
Using the logging library caused logging to produce the incorrect
line numbers. It looked pretty, but I am backing it out.


pkg/actor/decommission.go
```diff
@@ -98,7 +97,7 @@ func (d decommission) Act(ctx context.Context, cluster *resource.Cluster) error
 		log.Error(err, "We cannot decommission if there are less than 3 nodes", "nodes", nodes)
 		return err
 	}
-	log.V(int(zapcore.InfoLevel)).Info("replicas decommisioning", "status.CurrentReplicas", status.CurrentReplicas, "expected", cluster.Spec().Nodes)
+	log.Info("replicas decommisioning", "status.CurrentReplicas", status.CurrentReplicas, "expected", cluster.Spec().Nodes)
```
keith-mcclellan, line 100: why is this line a different format than 88 or 114?

pkg/actor/partitioned_update.go
```diff
@@ -202,7 +202,7 @@ func (up *partitionedUpdate) Act(ctx context.Context, cluster *resource.Cluster)
 	// TODO test downgrades
 	// see https://github.com/cockroachdb/cockroach-operator/issues/208
 
-	log.V(int(zapcore.InfoLevel)).Info("update starting with partitioned update", "old version", currentVersionCalFmtStr, "new version", versionWantedCalFmtStr, "image", containerWanted)
+	log.Info("update starting with partitioned update", "old version", currentVersionCalFmtStr, "new version", versionWantedCalFmtStr, "image", containerWanted)
```
keith-mcclellan, line 205: why isn't this line the same format as 163?

keith-mcclellan (changes requested):
I saw two formatting inconsistencies otherwise looks good

pkg/actor/decommission.go
chrislovecnm: Because you can call `log.Info` directly and it takes into consideration the logging level already.  You do not have to set the logging level by hand.

keith-mcclellan (approved):
LGTM

PR #523, opened 2021-05-25
Removing part of e2e

chrislovecnm:
Because we are now creating the cluster role for openshift testing
we cannot install the manifest, since it has the cluster role in
it.  I tried to do a delete and unfortunately the delete hangs.
Just removing it for now.


keith-mcclellan (approved):
LGTM

PR #524, opened 2021-05-25
Fixing GKE e2e testing role creation

chrislovecnm:
At some point in the testing the cluster role may exist but is
in the state of deleting.  I put in an exponential backoff in
to retry the creation of the cluster role and the cluster role
binding.


keith-mcclellan (approved):
LGTM

PR #526, opened 2021-05-26
Adding PreStop Hook - Removing LivenessProbe

chrislovecnm:
Adding the drain PreStop Hook for CRDB.
Removed LivenessProbe probe due to problems when pods are resource constrained.

### DIFF
--- a/pkg/resource/statefulset.go
+++ b/pkg/resource/statefulset.go
@@ -243,13 +243,16 @@ func (b StatefulSetBuilder) MakeContainers() []corev1.Container {
 			Name:            DbContainerName,
 			Image:           image,
 			ImagePullPolicy: *b.Spec().Image.PullPolicyName,
-			// Lifecycle: &corev1.Lifecycle{
-			// 	PreStop: &corev1.Handler{
-			// 		Exec: &corev1.ExecAction{
-			// 			Command: []string{"/cockroach/cockroach", "node", "drain", b.SecureMode()},
-			// 		},
-			// 	},
-			// },
+			Lifecycle: &corev1.Lifecycle{
+				PreStop: &corev1.Handler{
+					Exec: &corev1.ExecAction{
+						Command: []string{
+							"sh", "-c",
+							fmt.Sprintf("/cockroach/cockroach node drain %s || exit 0", b.SecureMode()),
+						},
+					},
+				},
+			},
 			Resources: b.Spec().Resources,
 			Command:   b.commandArgs(),
 			Env:       b.envVars(),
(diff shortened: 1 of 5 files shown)


keith-mcclellan (changes requested):
@chrisseto is recommending we eliminate the liveness probe as we've seen it cause nodes to restart unexpectedly in resource constrained situations (like when the DB is very busy)

pkg/resource/statefulset.go
```diff
@@ -247,6 +247,13 @@ func (b StatefulSetBuilder) MakeContainers() []corev1.Container {
 			Resources:       b.Spec().Resources,
 			Command:         b.commandArgs(),
 			Env:             b.envVars(),
+			Lifecycle: &corev1.Lifecycle{
+				PreStop: &corev1.Handler{
+					Exec: &corev1.ExecAction{
+						Command: []string{"/cockroach/cockroach", "node", "drain", b.SecureMode()},
```
chrisseto, line 253: @chrislovecnm have you gotten this to executre successfully? @alinadonisa and I couldn't get it to run for some reason 🤔 

I would also recommend wrapping this in `sh -c "%s || exit 0"`. If for whatever reason a lifecycle hook exits with a non-zero status code, the pod will be `SIGKILL`'d which is the last thing that we want if draining fails...

> If either a PostStart or PreStop hook fails, it kills the Container.

https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/#hook-handler-execution

chrislovecnm: what is the full string that I should be using?

chrislovecnm: Got it from @chrisseto on slack

chrislovecnm:
> @chrisseto is recommending we eliminate the liveness probe as we've seen it cause nodes to restart unexpectedly in resource constrained situations (like when the DB is very busy)

@keith-mcclellan fixed.  PTAL - The trade-off is that, for example, a pod that in a race condition will not get restarted.

keith-mcclellan:
> > @chrisseto is recommending we eliminate the liveness probe as we've seen it cause nodes to restart unexpectedly in resource constrained situations (like when the DB is very busy)
> 
> @keith-mcclellan fixed. PTAL - The trade-off is that, for example, a pod that in a race condition will not get restarted.

Luckily, we have other mechanisms in the database that will eject a node from the cluster (which automatically shuts the process down) in this scenario so k8s knows to rotate the pod.

chrisseto:
> The trade-off is that, for example, a pod that in a race condition will not get restarted.

Shouldn't that be gated by the readiness probe? CC has been running without a liveness probe for > 1 year now and hasn't had any issues.

chrisseto (approved):
LGTM

chrislovecnm:
> Shouldn't that be gated by the readiness probe? CC has been running without a liveness probe for > 1 year now and hasn't had any issues.

Nope, but as @keith-mcclellan said, CRDB has other mechanisms in place to eject nodes from the cluster.  Ready probe only runs to verify that a pod has started.  The Liveness probe runs periodically.

PR #527, opened 2021-05-27
Add Pod Health Checker

chrislovecnm:
Current implementation for all rolling restarts (regardless of cluster size):
This will be triggered by a rolling update on the statefulset, which implies that we are updating one pod at a time.

We wait on the k8s readiness (which means /health?ready=1) of ALL pods in the the cluster.

We check _status/vars on all cockroachdb pods looking for pairs like
ranges_underreplicated{store="1"} 0 and wait if any are non-zero until all are 0. We can recheck every 10 seconds.

We wait 22 seconds and we check again _status/vars on all cockroachdb pods looking for pairs like
ranges_underreplicated{store="1"} 0.

Once this checks complete, we update another pod.

Integration of the implementation on rolling restarts and upgrades.


keith-mcclellan (approved):
LGTM

PR #528, opened 2021-05-27
Removing unused scripts and fixing create script

chrislovecnm:
Removing scripts that we are not using and fixing gke cluster
create script.


udnay (approved):
LGTM

Issue #193, opened 2020-10-12
run e2e tests in parallel and refactor e2e tests


chrislovecnm:
The e2e tests do not seem to run in parallel and are already taking 8 minutes to write.  The test suite that we are using does not respect the -test.parallel flag and I am uncertain why.

johnrk:
@keith-mcclellan to follow up with @chrislovecnm on this.

chrislovecnm:
We need to refactor the e2e tests into separate go files. Each go test file needs to be in a separate package.

PR #530, opened 2021-05-27
Refactoring E2E

chrislovecnm:
- Moving code around to encapsulate our e2e code under testutil.
- Creating multiple different go files that contain our various
e2e tests.
- Updated Makefile to allow for multiple tests.
- Created new tests to run with GitHub CI.

Closes https://github.com/cockroachdb/cockroach-operator/issues/193


udnay (dismissed):
LGTM

pkg/update/update_cockroach_version_common.go
```diff
@@ -86,7 +86,7 @@ func makeIsCRBPodIsRunningNewVersionFunction(
 			return statusError
 		} else if err != nil { // this is an error
 			// TODO uncertain why this is logging is throwing an error
-			l.Error(err, "error getting pod")
+			// l.Error(err, "error getting pod")
```
chrislovecnm, line 89: I commented this out in another PR and the merge must have gotten messed up.  https://github.com/cockroachdb/cockroach-operator/issues/532 is open to deal with this.

Makefile
```diff
@@ -91,11 +92,11 @@ test/e2e/testrunner-kind:
 # test/e2e/testrunner-kind.
 # After the tests run the cluster is deleted.
 # If you need a unique cluster name override CLUSTER_NAME.
-.PHONY: test/e2e/kind
```
chrislovecnm, line 94: Anyone know if we can do a .PHONY while using %?

keith-mcclellan:
Since we're updating the testing framework, should we bump to the latest release of Bazel at the same time? 

https://github.com/bazelbuild/bazel/releases/tag/4.1.0

e2e/decomission/decomission_test.go
```diff
@@ -0,0 +1,97 @@
...
+	code := testenv.RunCode(m, e)
+	os.Exit(code)
+}
+
+func TestDecommissionFunctionality(t *testing.T) {
+
+	// Testing removing and decommisioning a node.  We start at 4 node and then
+	// remove the 4th node
+
+	if testing.Short() {
```
keith-mcclellan, line 59: if we're breaking this up, shouldn't this be caught before we ever get this far? I don't think the decommission test does anything in short mode

chrislovecnm:
> Since we're updating the testing framework, should we bump to the latest release of Bazel at the same time @chrislovecnm?
> 
> https://github.com/bazelbuild/bazel/releases/tag/4.1.0

In another PR ...

e2e/upgrades/upgrades_test.go
```diff
@@ -0,0 +1,200 @@
...
+		},
+	}
+
+	steps.Run(t)
+}
+
+func TestUpgradesMajorVersion20to21(t *testing.T) {
+
+	// We are doing a major version upgrade here
+	// 19 to 20
```
keith-mcclellan, line 105: we need to update this comment

pkg/testutil/require.go
```diff
@@ -222,7 +224,9 @@ func statefulSetIsReady(ss *appsv1.StatefulSet) bool {
 	return ss.Status.ReadyReplicas == ss.Status.Replicas
 }
 
-func requireDownGradeOptionSet(t *testing.T, sb testenv.DiffingSandbox, b testutil.ClusterBuilder, version string) {
+// TODO we are not using this
```
keith-mcclellan, line 227: what do you mean we aren't using this? We aren't checking to see if the downgrade option is set before starting an upgrade?

pkg/testutil/require.go
```diff
@@ -261,7 +265,11 @@ func requireDownGradeOptionSet(t *testing.T, sb testenv.DiffingSandbox, b testut
 	}
 
 }
-func requireDecommissionNode(t *testing.T, sb testenv.DiffingSandbox, b testutil.ClusterBuilder, numNodes int32) {
+
+// TODO I do not think this is correct.  Keith mentioned we need to check something else.
```
keith-mcclellan, line 269: we need to check `cockroach node list` to see if the decommissioned node is listed as "decommissioned" and not as something else (like down or failed)

keith-mcclellan (changes requested):

e2e/decomission/decomission_test.go
chrislovecnm: You can still run bazel test //e2e/... and then choose which tests to skip.  I use the skip to isolate a test because I can never remember the full option to just run one test name in go. So for developer friendliness can we leave it in there.

pkg/testutil/require.go
chrislovecnm: We are checking it in the operator code, but we are not using this code in any of the e2e tests.

pkg/testutil/require.go
chrislovecnm: Sure, but not in this PR.  Please file an issue

Makefile
chrislovecnm: @chrisseto ??

chrisseto: Seems like no?

https://stackoverflow.com/questions/65664156/how-can-you-make-a-wildcard-make-rule-phony

keith-mcclellan:
> > Since we're updating the testing framework, should we bump to the latest release of Bazel at the same time @chrislovecnm?
> > https://github.com/bazelbuild/bazel/releases/tag/4.1.0
> 
> In another PR ...

opened https://github.com/cockroachdb/cockroach-operator/issues/534 for this so we don't forget

pkg/testutil/require.go
keith-mcclellan: https://github.com/cockroachdb/cockroach-operator/issues/535

Makefile
keith-mcclellan: thats a bummer. lets go back to the 70s and yell at Stuart Feldman at Bell Labs.

pkg/testutil/require.go
keith-mcclellan: https://github.com/cockroachdb/cockroach-operator/issues/537

keith-mcclellan (approved):
LGTM

PR #540, opened 2021-05-28
Drop 19.2.x support, centralize supported crdb versions, fix sed expression limit

rail:
* Drop 19.2.12 support.
* Centralize list of supported CRDB versions used by some scripts


keith-mcclellan (approved):
LGTM

hack/update-pkg-manifest.sh
```diff
@@ -69,7 +68,7 @@ rm -rf ${DEPLOY_PATH}/${RH_BUNDLE_VERSION}
 "$kstomize" build config/manifests | "$opsdk" generate packagemanifests -q --version ${RH_BUNDLE_VERSION} ${RH_PKG_MAN_OPTS} --output-dir ${DEPLOY_PATH} --input-dir ${DEPLOY_PATH} --verbose
 # cat ${DEPLOY_PATH}/${RH_BUNDLE_VERSION}/cockroach-operator.clusterserviceversion.yaml | sed -e "s+RH_COCKROACH_OP_IMAGE_PLACEHOLDER+${RH_COCKROACH_OP_IMG}+g" -e "s+RH_COCKROACH_DB_IMAGE_PLACEHOLDER+${RH_COCKROACH_DATABASE_IMAGE}+g" -e "s+CREATED_AT_PLACEHOLDER+"$(date +"%FT%H:%M:%SZ")"+g"> ${DEPLOY_PATH}/${RH_BUNDLE_VERSION}/csv.yaml
 cat ${DEPLOY_PATH}/${RH_BUNDLE_VERSION}/cockroach-operator.clusterserviceversion.yaml | sed "s+RH_COCKROACH_OP_IMAGE_PLACEHOLDER+${RH_COCKROACH_OP_IMG}+g; s+CREATED_AT_PLACEHOLDER+"$(date +"%FT%H:%M:%SZ")"+g"> ${DEPLOY_PATH}/${RH_BUNDLE_VERSION}/csv.yaml
-for v in "${SUPPORTED_VERSIONS[@]}"
+for v in $(grep -v '#' ${REPO_ROOT}/crdb-versions.txt)
```
chrislovecnm, line 71: Does this fail if the file does not exist?

chrislovecnm (commented):
Error handling question

hack/update-pkg-manifest.sh
rail: It should fail, because the pipe exits non zero.

chrislovecnm: Cool. How difficult is a regex to test the numbering?  Do we need to do a follow-up PR?

rail: I'd say, replace the script with something not bash (go!) and have it use calver to sort things. And we can unittest things. I'll file an issue.

chrislovecnm (approved):
LGTM

PR #544, opened 2021-05-28
Split CI

rail:
Splitting CI into separate files makes it possible to rerun a single
workflow without rerunning all at once.


chrislovecnm (approved):
LGTM

PR #547, opened 2021-05-28
Openshift Scripting E2E Testing

chrislovecnm:
## Why we need this

We are not testing the packaging and installation of the operator on OpenShift
during out nightly builds.  We have pushed a release to the RH marketplace where the packaging was bad and
the operator installed.  This should help guarantee that our operator can at least:

1. be installed on OpenShift
2. start
3. install a basic CRDB database (uses example.yaml)

## Work included

- fixed bazel targets to build and push OpenShift packaging containers
- created make targets to run above
- created e2e test that automatically installs the operator on OpenShift
- created bash scripts that allow for running all of the complex stuff above
- added documentation to the Makefile and the tests on how to run all of this

## TODO

- [x] check bash script error handling
- [x] add instructions on running
- [x] download oc
- [x] rename push sript and remove unused script


hack/push-openshift-images.sh
```diff
@@ -0,0 +1,49 @@
...
+echo ${REPO_ROOT}
+[[ -z "$APP_VERSION" ]] && { echo "Error: APP_VERSION not set"; exit 1; }
+[[ -z "$DOCKER_REGISTRY" ]] && { echo "Error: DOCKER_REGISTRY not set"; exit 1; }
+
+echo "Running with args APP_VERSION=$APP_VERSION DOCKER_REGISTRY=$DOCKER_REGISTRY"
+
+cd deploy/certified-metadata-bundle/cockroach-operator/
+
+docker build -t ${DOCKER_REGISTRY}/cockroachdb-operator-bundle:${APP_VERSION}  -f bundle.Dockerfile .
+docker build -t  ${DOCKER_REGISTRY}/cockroachdb-operator-bundle:${APP_VERSION} -f  bundle.Dockerfile .
```
chrislovecnm, line 47: Bug here.  I have the same commands

e2e/openshift/openshift_packaging_test.go
```diff
@@ -0,0 +1,268 @@
...
+			"crdb",
+			"test",
+			800*time.Second,
+			10*time.Second,
+		),
+		"crdb database did not start",
+	)
+}
+
+func writeTempFile(t *testing.T, b []byte) string {
```
chrislovecnm, line 178: I cannot use temp files because they are not available to the subprocess when I shell out.

Makefile
```diff
@@ -284,22 +291,21 @@ release/bundle-image:
...
+	APP_VERSION=$(APP_VERSION) \
+	hack/cleanup-packaging.sh 
+
+.PHONY: test/clean-up
+test/clean-up:
+	VERSION=$(VERSION) \
+	APP_VERSION=$(APP_VERSION) \
+	hack/cleanup-packaging.sh 
 
 CHANNELS?=beta,stable
```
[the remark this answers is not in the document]
chrislovecnm: Sure, not in this PR. And i dont fullu know how @rail does this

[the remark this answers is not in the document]
chrislovecnm: Opened an issue

Makefile
```diff
@@ -223,7 +223,7 @@ release/versionbump:
 
 .PHONY: release/image
 release/image:
-	bazel clean --expunge
+	# bazel clean --expunge
```
keith-mcclellan, line 226: We added this because otherwise we ended up with a stale base image.  Have we resolved that issue another way?

chrislovecnm: We have the correct stamp in the Workspace file, so the stamp will update it now

hack/push-openshift-images.sh
```diff
@@ -0,0 +1,60 @@
...
+cd deploy/certified-metadata-bundle/cockroach-operator/
+docker build -t  ${DOCKER_REGISTRY}/cockroachdb-operator-bundle:${APP_VERSION} -f  bundle.Dockerfile .
+docker push ${DOCKER_REGISTRY}/cockroachdb-operator-bundle:${APP_VERSION}
+
+cd -
+echo "Running opm index with args APP_VERSION=$APP_VERSION DOCKER_REGISTRY=$DOCKER_REGISTRY"
+"$opm" index add --bundles  ${DOCKER_REGISTRY}/cockroachdb-operator-bundle:${APP_VERSION} \
+  --tag  ${DOCKER_REGISTRY}/cockroachdb-operator-index:${APP_VERSION} -c docker
+
+docker build -t ${DOCKER_REGISTRY}/cockroachdb-operator-index:${APP_VERSION}  -f bundle.Dockerfile .
```
alinadonisa, line 59: @chrislovecnm please remove  L59. In my scripts I was doing again a build on the bundle not index that has non sense

chrislovecnm:
@rail I doing some stuff here that might impact release so PTAL

@keith-mcclellan / @udnay this is a pretty complex testing PR, so I would appreciate your eyes on it.

Makefile
```diff
@@ -19,7 +19,7 @@
 
 # values used in workspace-status.sh
 DOCKER_REGISTRY?=cockroachdb
-DOCKER_IMAGE_REPOSITORY?=cockroachdb-operator
+DOCKER_IMAGE_REPOSITORY?=cockroach-operator
```
chrislovecnm, line 22: @rail this was WRONG.  Is this used in the release process?

Makefile
```diff
@@ -254,7 +264,9 @@ RH_BUNDLE_VERSION?=$(VERSION)
 RH_DEPLOY_PATH="deploy/certified-metadata-bundle"
 RH_DEPLOY_FULL_PATH="$(RH_DEPLOY_PATH)/cockroach-operator/"
 RH_COCKROACH_DATABASE_IMAGE=registry.connect.redhat.com/cockroachdb/cockroach:$(COCKROACH_DATABASE_VERSION)
-RH_OPERATOR_IMAGE?=registry.connect.redhat.com/cockroachdb/cockroachdb-operator:$(APP_VERSION)
+# TODO this is coackroachdb-operator, but we are using coachroach-operator
```
chrislovecnm, line 267: @rail here is another thing that was wrong.  Are we overriding cockroachdb -> cockroach during the release?

Makefile
```diff
@@ -297,22 +308,35 @@ release/bundle-image:
...
+	VERSION=$(VERSION) \
+	hack/cleanup-packaging.sh 
+
+# This target pushes the OpenShift bundle, then uses opm to push the index bundle.
+.PHONY: test/push-openshift-images
+test/push-openshift-images:
+	APP_VERSION=$(APP_VERSION) \
+	DOCKER_REGISTRY=$(DOCKER_REGISTRY) \
 	bazel run --stamp --platforms=@io_bazel_rules_go//go/toolchain:linux_amd64 \
-		//hack:opm-build-index $(OLM_REPO) $(OLM_BUNDLE_REPO) $(TAG) $(RH_BUNDLE_VERSION) $(RH_COCKROACH_DATABASE_IMAGE)
```
chrislovecnm, line 315: The opm-build-index bazel target was a dev target that never worked.  I have replaced it with a target that works, and could do the release for us as well.

Makefile
chrislovecnm: @rail can you please confirm this

Makefile
keith-mcclellan: This is actually correct here - in redhat it's 'cockroachdb-operator' in github and dockerhub it's 'cockroach-operator'

Makefile
```diff
@@ -236,7 +247,6 @@ release/gen-files: release/gen-templates
 
 .PHONY: release/image
 release/image:
-	bazel clean --expunge
```
keith-mcclellan, line 239: We put this in to stop us from using a stale UBI base image - how are we making sure that no longer happens?

keith-mcclellan (changes requested):

Makefile
rail: The release script sets it to [cockroachdb/cockroach-operator](https://github.com/cockroachdb/cockroach-operator/blob/a65d8e9dcb100035010bbd8ec294529bb29837a7/build/release/teamcity-publish-release.sh#L35)

Makefile
rail: Ideally we shouldn't run this target manually. For rare cases, when we have to, I'd leave the clean step.

Makefile
rail: `registry.connect.redhat.com/cockroachdb/cockroachdb-operator:$(APP_VERSION)` is correct. Not sure if we want to rename it...

Makefile
chrislovecnm: Yah we should have that in the Makefile, I don’t like stuff spread all over the place 😂

Makefile
udnay: We should pick one, can dockerhub do redirects?

e2e/openshift/openshift_packaging_test.go
```diff
@@ -0,0 +1,340 @@
...
+
+	// create a new project/namespace for testing
+	args = []string{
+		"new-project",
+		"test",
+	}
+
+	// we cannot do a require.NoError on process, because oc tries to modify
+	// the kubeconfig file, and with bazel you cannot do this
+	process.ExecJUnit("oc", args, os.Environ())
```
udnay, line 157: So I understand, the error thrown here will be bazel complaining that `kubeconfig` was modified? Is this the case for the other junit execs?

e2e/openshift/openshift_packaging_test.go
```diff
@@ -0,0 +1,340 @@
...
+	}
+	return string(b)
+}
+
+// wriiteFile writes out b byte array to a random file in the
+// os.TempDir and returns the name of that file.
+func writeFile(t *testing.T, b []byte) string {
+	rand.Seed(time.Now().UnixNano())
+	fileName := fmt.Sprintf("%s-test.yaml", randSeq(10))
+	fileName = filepath.Join(os.TempDir(), fileName)
```
udnay, line 249: Did you want to use `t.TempDir()` instead so that there's no artifacts left behind?

e2e/openshift/openshift_packaging_test.go
```diff
@@ -0,0 +1,340 @@
...
+	b.WriteString(operatorGroup)
+
+	operatorFilename := writeFile(t, b.Bytes())
+
+	files := []string{catalogFilename, subFilename, operatorFilename}
+
+	// install YAML files which install the operator
+	for _, f := range files {
+		fmt.Println(f)
+		args := []string{"delete", "-f", f}
```
udnay, line 182: The files all have a `nonce` suffix, I doubt this will make a difference, maybe a `defer` to clean up at the end?

udnay (commented):
So this installs the operator in the non-default namespace and makes sure a cluster comes up in openshift?

Makefile
chrislovecnm: There can only be one! Forgive the movie reference 😂. Can we just get one name please

Makefile
chrislovecnm: The clean step kills development. We cannot have it here. If we need a release target I understand that.

chrislovecnm:
> So this installs the operator in the non-default namespace and makes sure a cluster comes up in openshift?

Yep!

e2e/openshift/openshift_packaging_test.go
chrislovecnm: If we fail with a err will the defer still run in the test framework?

e2e/openshift/openshift_packaging_test.go
chrislovecnm: Ohhhhh! Perfect. Did not know that trick!

Makefile
keith-mcclellan: we can't change the Red Hat name without breaking our Red Hat Marketplace listing

e2e/openshift/openshift_packaging_test.go
chrislovecnm: oc tries to change the kubeconfig file and bazel does not allow it to, since the file is outside of the sandbox. So oc throws an error and the test fails. The project is created successfully. So the oc command does what we needed it to do. We could try to create a namespace rather than creating a new project, but I do not know if that would work. Oh we could try to copy the kubeconfig into the sandbox.

The next steps in the test will fail if the ns is missing.

Makefile
rail: We can add this step to the release script then.

Makefile
rail: I'm not quite sure how to confirm this... What did exactly change?

Makefile
chrislovecnm: @rail are we running the release locally or on a clean vm/test?  We may not have the cache during the release

Makefile
chrislovecnm: Yeah we do not a checksum/digest in the WORKSPACE file, so I will add this back in, but it kills dev time ;(. Will file an issue.

Makefile
chrislovecnm: I reverted and I will add a build issue

chrislovecnm: I reverted the change and added an issue to fix the container pull.

Makefile
rail: I think the issue is [this thing](https://github.com/cockroachdb/cockroach-operator/blob/a65d8e9dcb100035010bbd8ec294529bb29837a7/cmd/cockroach-operator/BUILD.bazel#L37-L45), which is not deterministic. I'm not sure how to make it deterministic. :/

Makefile
```diff
@@ -254,7 +267,7 @@ RH_BUNDLE_VERSION?=$(VERSION)
 RH_DEPLOY_PATH="deploy/certified-metadata-bundle"
 RH_DEPLOY_FULL_PATH="$(RH_DEPLOY_PATH)/cockroach-operator/"
 RH_COCKROACH_DATABASE_IMAGE=registry.connect.redhat.com/cockroachdb/cockroach:$(COCKROACH_DATABASE_VERSION)
-RH_OPERATOR_IMAGE?=registry.connect.redhat.com/cockroachdb/cockroachdb-operator:$(APP_VERSION)
+RH_OPERATOR_IMAGE?=registry.connect.redhat.com/cockroachdb/cockroachb-operator:$(APP_VERSION)
```
udnay, line 270: spelling missing `d`

chrislovecnm: 😆 opps

Makefile
```diff
@@ -297,22 +309,35 @@ release/bundle-image:
...
-dev/opm-build-index:
-	RH_BUNDLE_REGISTRY=$(RH_BUNDLE_REGISTRY) \
-	RH_BUNDLE_IMAGE_REPOSITORY=$(RH_BUNDLE_IMAGE_REPOSITORY) \
-	RH_BUNDLE_VERSION=$(RH_BUNDLE_VERSION) \
-	RH_DEPLOY_PATH=$(RH_DEPLOY_FULL_PATH) \
+# APP_VERSION
+# VERSION
+# RH_BUNDLE_VERSION
+# RH_OPERATOR_IMAGE
+# DOCKER_REGISTRY
```
udnay, line 326: Where are these set?

udnay (approved):

Makefile
chrislovecnm: We have values in the Makefile that are defaults. Otherwise you can override them on the command line. You can take a look at my script that I am using to run the test.

keith-mcclellan (commented):
this LGTM - @udnay did you want other changes?

keith-mcclellan (approved):

Issue #542, opened 2021-05-28
[Regression] Decommission is broken


keith-mcclellan:
The statefulset is removing the pod before `cockroach node decommission` is being executed, so the node is showing up as failed instead of as decommissioned.

keith-mcclellan:
the PVC gets deleted as well so there is no way to recover from this state.

alinadonisa:
> the PVC gets deleted as well so there is no way to recover from this state.

@keith-mcclellan  so  we keep the  PVC, and we delete them only if the decommission command it is successfull?
@chrisseto   I moved the Prune after after the downscale code. Now the reason that the decommission gave an error is still to be determined.
```
// Before doing any scaling, prune any PVCs that are not currently in use.
	// This only needs to be done when scaling up but the operation is a noop
	// if there are no PVCs not currently in use.
	// As of v20.2.0, CRDB nodes may not be recommissioned. To account for
	// this, PVCs must be removed (pruned) before scaling up to avoid reusing a
	// previously decommissioned node.
	// Prune MUST be called before scaling as older clusters may have dangling
	// PVCs.
	// All underlying PVs and the storageclasses they were created with should
	// make use of reclaim policy = delete. A reclaim policy of retain is fine
	// but will result in wasted money, recycle should be considered unsafe and
	// is officially deprecated by kubernetes.
	if err := s.PVCPruner.Prune(ctx); err != nil {
		return errors.Wrap(err, "initial PVC pruning")
	}
```

keith-mcclellan:
Decommission should run as follows functionally:

1) Validate that the node count after decommission is still >= 3 (node decommissions CAN be run in parallel)
2) Run `cockroach node decommission` 
3) If fails, annotate the CR to stop further runs without user input, log an error, and reset the node count to the original amount
4) If successful, wait 60 seconds and run `cockroach node status --decommission` (or optionally `cockroach node status --all`) to validate that the node is decommissioned and the database is ready for it to exit the cluster
5) If `cockroach node status --decommission` does not show the node as decommissioned, do same as #3
6) Stop the decommissioned pod
7) Run health checker to validate that we have 0 under-replicated ranges using the same pattern as a rolling restart
8) Delete the SS and PVC

Decommission tests:

Positive case 1 -
After `cockroach node decommission` is run, health-checker should show 0 under-replicated ranges AND `cockroach node status --decommission` should show the node as decommissioned.

Positive case 2 -
[...]

PR #552, opened 2021-05-31
Upgrading using cockroachDBVersion field and image in format sha256 on RELATED_IMAGE fails

alinadonisa:
Upgrading using cockroachDBVersion on openshift fails. See issue: https://github.com/cockroachdb/cockroach-operator/issues/542.

 Things that were done:
- I  modified test framework to check image from cockroackDBVersion field on upgrade
- I added e2e test suite that use cockroackDBVersion as input on upgrades. I also used in the tests  the RELATED_IMAGES env var images in sha256 format to emulate the Openshift CI
- I added a fix for this bug
- I added a unit test for our scenario


e2e/upgrades/upgrades_test.go
```diff
@@ -100,6 +100,59 @@ func TestUpgradesMinorVersion(t *testing.T) {
...
+		t.Skip("skipping test in short mode.")
+	}
+
+	testLog := zapr.NewLogger(zaptest.NewLogger(t))
+
+	actor.Log = testLog
+
+	sb := testenv.NewDiffingSandbox(t, env)
+	sb.StartManager(t, controller.InitClusterReconcilerWithLogger(testLog))
+	os.Setenv("RELATED_IMAGE_COCKROACH_v21_1_1", "cockroachdb/cockroach@sha256:7c84559a33db90b52f8179c904818525e45852b683bd6272f61dcf54c103f5b1")
```
udnay, line 123: Does this need to be against redhat specifically?

udnay: I'm trying to run this test and get
```Back-off pulling image registry.connect.redhat.com/cockroachdb/cockroach:v20.2.10
job image incorrect
``` 
What am I doing wrong?

udnay: See you pushed a new commit nevermind

e2e/upgrades/upgrades_test.go
```diff
@@ -100,6 +100,59 @@ func TestUpgradesMinorVersion(t *testing.T) {
...
+	os.Setenv("RELATED_IMAGE_COCKROACH_v20_2_10", "cockroachdb/cockroach@sha256:a1ef571ff3b47b395084d2f29abbc7706be36a826a618a794697d90a03615ada")
+	builder := testutil.NewBuilder("crdb").WithNodeCount(3).WithTLS().
```
udnay, line 125: Can this just be a 1 node cluster? We're testing that the upgrade happens, not that the rolling restart functionality works right?

alinadonisa: @udnay actually it is a 3 nodes cluster... I already fixed it. It was a copy of the test for upgrades using the image field... I fixed it also there. The description of the test says 1 node but it is a 3 nodes cluster.

udnay: Right, so what is this test actually trying to validate? The image string?

pkg/actor/partitioned_update.go
```diff
@@ -260,6 +260,10 @@ func getStsAnnotation(statefulSet *appsv1.StatefulSet, key string) string {
 }
 
 func getImageNameNoVersion(image string) string {
+	// if somehow this arrives as sha256 we do not extract version
```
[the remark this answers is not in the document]
alinadonisa: @udnay  I added some unit tests, but there should not be any negative cases here, because the image it is validated by the version checker job first and it is added by our workflow.  It was a bug in our code because I did not treated the case of Openshift where they replace the supported images with the sha256.

e2e/upgrades/upgrades_test.go
alinadonisa: This test should validate that we can upgrade a cluster using the cockroachDBversion field instead of the image field, using the supported versions env var that we declare in the RELATED_IMAGES env. vars. Actually it is the scenario that Keith described in the issue.  Without the fix I added this test would finish with a broken upgrade.

pkg/actor/partitioned_update.go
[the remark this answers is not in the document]
udnay: Then why not put that code into the version checker if it can already do parsing and validating of the versions?

e2e/upgrades/upgrades_test.go
udnay: Can you make that more explicit in the test itself so people understand the intent of this test?

udnay:
This is looking good, thank @alinadonisa for pushing it through

e2e/upgrades/upgrades_test.go
```diff
@@ -100,6 +100,59 @@ func TestUpgradesMinorVersion(t *testing.T) {
 	steps.Run(t)
 }
 
+// TestUpgradesMinorVersion tests a minor version bump
+func TestUpgradesMajorVersion20to21WithCockroachDbVersion(t *testing.T) {
```
chrislovecnm, line 104: Can we add this to another pkg? The upgrade tests are taking time :(

e2e/upgrades/upgrades_test.go
```diff
@@ -100,6 +100,59 @@ func TestUpgradesMinorVersion(t *testing.T) {
...
+
+	// We are testing a Minor Version Upgrade with
+	// partition update
+	// Going from v20.2.10 to v21.1.0
```
chrislovecnm, line 108: Is this a minor or major version upgrade?

pkg/update/update_cockroach_version.go
```diff
@@ -98,8 +99,10 @@ func UpdateClusterCockroachVersion(
 			return errors.Wrapf(err, "setting downgrade option for major roll forward failed")
 		}
 	}
-
-	wantImage := fmt.Sprintf("%s:%s", update.WantImageName, update.WantVersion.Original())
+	var wantImage string = update.WantImageName
```
chrislovecnm, line 102: Please add comments on what we are doing here

pkg/actor/partitioned_update.go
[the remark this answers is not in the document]
alinadonisa: The validation consist in running a job with the image the user sets. First step is checking that the job is running. If the job is running the image exists. Next we check to see if the image is a cockroachdb image by running an exec on the pod and trying to obtain the version tag. If also  this step succeeds we consider it a valid image.

pkg/actor/partitioned_update_test.go
```diff
@@ -40,3 +42,30 @@ func TestDeployedInCluster(t *testing.T) {
...
+	}
+	tests := []testCase{
+		{
+			Name:              "should_keep_the_sha256_format_complete",
+			ImageName:         "cockroachdb/cockroach@sha256:a1ef571ff3b47b395084d2f29abbc7706be36a826a618a794697d90a03615ada",
+			ExpectedImageName: "cockroachdb/cockroach@sha256:a1ef571ff3b47b395084d2f29abbc7706be36a826a618a794697d90a03615ada",
+		},
+		{
+			Name:              "should_remove_the_version_for_non_sha256_format_of_the_image",
+			ImageName:         "cockroachdb/cockroach:v20.2.0",
```
keith-mcclellan, line 60: I don't think this does what we want.  Customers should be able to use SHAs OR tags - a customer is probably going to manually set a tag, and the cockroachDBVersion API is always going to tag a SHA.  Unless I'm missing something here?

e2e/upgrades/upgrades_test.go
alinadonisa: I added some more comments, but please check again and tell me if you need also other changes

e2e/upgrades/upgrades_test.go
keith-mcclellan: I would prefer to have identical test suites, one that runs the upgrade tests using image.name and one using cockroachDBversion so we can do this in parallel.  Thoughts?

keith-mcclellan (changes requested):

e2e/upgrades/upgrades_test.go
alinadonisa: @keith-mcclellan I agree with you, because the same scenario must work also with api field

e2e/upgrades/upgrades_test.go
alinadonisa: fixed :)

pkg/actor/partitioned_update_test.go
alinadonisa: @Keith I will rephrase... the sha256 format was for the RELATED_IMAGE that we add in the manifest, and which Openshift CI replaces them with sha256.

keith-mcclellan: okay maybe I misunderstood what the test was doing then

pkg/update/update_cockroach_version.go
alinadonisa: done

e2e/upgrades/upgrades_test.go
alinadonisa: @keith-mcclellan , @chrislovecnm , @udnay  I added in another package the tests with the related image in the sha256 format

udnay:
Can this be rebased and merged ASAP as well?

alinadonisa:
> Can this be rebased and merged ASAP as well?

@udnay  yes ...:)

udnay (approved):
LGTM

keith-mcclellan (approved):
LGTM

PR #553, opened 2021-05-31
Prune after decommission of the node

alinadonisa:
Decommission is broken. See issue https://github.com/cockroachdb/cockroach-operator/issues/542


udnay:
Please add some more detail as to how the current change fixes the issue. Also can you please create a test that fails without this code change?

udnay (changes requested):
Please create a test case that would catch this error.

chrisseto (changes requested):
There's already a call to `.Prune` at the end of `EnsureScale`?

https://github.com/cockroachdb/cockroach-operator/pull/553/files#diff-7e9ec543c7f95e6e8422692269f9086076669624f3c0542a83d6ccf60cb8b268R157-R175

I don't see how this would change or fix anything besides making the operator vulnerable to the bug described in the comment that you've moved.

chrislovecnm:
> Please create a test case that would catch this error.

@udnay I would like a broken e2e test in another PR.

pkg/scale/scale.go
```diff
@@ -100,6 +84,21 @@ func (s *Scaler) EnsureScale(ctx context.Context, scale uint) error {
 			return err
 		}
 	}
+	// Before doing any scaling, prune any PVCs that are not currently in use.
+	// This only needs to be done when scaling up but the operation is a noop
+	// if there are no PVCs not currently in use.
+	// As of v20.2.0, CRDB nodes may not be recommissioned. To account for
```
keith-mcclellan, line 90: I don't know where this is coming from, but we can recommission nodes. see https://www.cockroachlabs.com/docs/v21.1/remove-nodes#step-2-recommission-the-decommissioning-nodes

The node won't automatically recommission, there's a step after the node starts before it's allowed to rejoin the cluster.  And if the node was properly decommissioned there's no need to recommission, a fresh PVC is preferred. but If the decommission didn't complete successfully, recommissioning is the best option to make the cluster healthy again.

keith-mcclellan (changes requested):

pkg/scale/scale.go
chrisseto: Nodes that have been fully decommissioned may not be recommissioned. Agreed that if the decommissioning process fails for whatever reason, recommissioning is the best option. However, if it succeeds the PVC can't be reused.

chrislovecnm:
We are closing this.

PR #555, opened 2021-06-01
Shutting off Decommission Actor by default

chrislovecnm:
We are having issues with the Decommission Actor so we are turning
it off by default.  You can enable the Actor via the feature flag.

The test case for this PR is that Decommission will run when we enable the flag.  Otherwise, it will not run.

### DIFF
--- a/e2e/decomission/decomission_test.go
+++ b/e2e/decomission/decomission_test.go
@@ -26,6 +26,7 @@ import (
 	"github.com/cockroachdb/cockroach-operator/pkg/controller"
 	"github.com/cockroachdb/cockroach-operator/pkg/testutil"
 	testenv "github.com/cockroachdb/cockroach-operator/pkg/testutil/env"
+	"github.com/cockroachdb/cockroach-operator/pkg/utilfeature"
 	"github.com/go-logr/zapr"
 	"github.com/stretchr/testify/require"
 	"go.uber.org/zap/zaptest"
@@ -35,9 +36,6 @@ import (
 // deployed in the cluster helps
 // We may have a threadsafe problem where one test starts messing with another test
 var parallel = *flag.Bool("parallel", false, "run tests in parallel")
-
-// run the pvc test
-var pvc = flag.Bool("pvc", false, "run pvc test")
 
 // TODO should we make this an atomic that is created by evn pkg?
 var env *testenv.ActiveEnv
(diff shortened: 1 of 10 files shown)


pkg/actor/decommission_test.go
```diff
@@ -0,0 +1,47 @@
...
+		WithUID("cockroachdb-uid").
+		WithPVDataStore("1Gi", "standard" /* default storage class in KIND */).
+		WithNodeCount(1).Cluster()
+	cluster.SetTrue(api.InitializedCondition)
+
+	scheme := testutil.InitScheme(t)
+	client := testutil.NewFakeClient(scheme)
+	deco := newDecommission(scheme, client, nil)
+
+	require.False(t, deco.Handles(cluster.Status().Conditions))
```
udnay, line 45: Can you also add a test to make sure that `Handles` returns true when the flag is enabled?

chrislovecnm: 😆 just pushed it ... great minds think alike

udnay (commented):
Looks good, one comment. Also where do we set feature flags to `true` for our `e2e` tests now that it's disabled by default?

e2e/decomission/decomission_test.go
```diff
@@ -51,10 +49,13 @@ func TestMain(m *testing.M) {
 	os.Exit(code)
 }
 
+// TestDecomissionFunctionality creates a cluster of 4 nodes and then decomissions on of the CRDB nodes.
+// It then checks that the cluster is stable and that decomissioning is successful.
 func TestDecommissionFunctionality(t *testing.T) {
 
 	// Testing removing and decommisioning a node.  We start at 4 node and then
 	// remove the 4th node
+	utilfeature.DefaultMutableFeatureGate.Set("UseDecommission=true")
```
chrislovecnm, line 58: @udnay we turn on Decommission here.

udnay: Ah, I missed it. Thought it was unit-tests.

udnay: Actually do any of the other `e2e` tests assume decommission is on?

chrislovecnm: @udnay no they do not.  We have had the decommission actor cause problems in other e2e tests, so I could enable it everywhere.  What do you think?

keith-mcclellan (commented):
What happens when we reduce the node count (say from 4 to 3) if this actor is disabled?  Do we just not decrease the statefulset count?

udnay (approved):
LGTM

PR #557, opened 2021-06-02
Fix e2e test for failing decommission actor

alinadonisa:
### [Regression] Decommission is broken:

The statefulset is stopping the pod before cockroach node decommission is being executed, so the node is showing up as failed instead of as decommissioned.

The PVC also gets deleted so there is no way to recover from this state if the scale down caused a loss of quorum because the data gets destroyed.

### Investigation

1. The e2e test for decommission was passing. Why? 

We considered a success that the statefulset is ready and the number of nodes from the spec is the same as the statefulset replicas. Also we check that the database connection. Fact that was true in our scenario.

```golang
        testutil.RequireClusterToBeReadyEventuallyTimeout(t, sb, builder, 500*time.Second)
        testutil.RequireDecommissionNode(t, sb, builder, 3)
        testutil.RequireDatabaseToFunction(t, sb, builder)
```

2. How can we fix it?

  - **make the test fail**:  @keith-mcclellan  added the missing piece:

The e2e test for decommission is incomplete. After checking the statefulset replicas  we need to run also command:
```
cockroach node status --decommission --certs-dir=certs --host=<address of any live node>
```
 to check that the decommission was successful.

3. Hypothesis:
-  the decommission code is code taken from **Intrusion**, so it should be more than tested. :) 
-  in the logs we see the error and than the other actors, including **partial update**(upgrade) actor is running that could alter the statefulset replicas.

4. What did we changed and could have an impact? 
-  the order of the actors, now decommission runs first?

### Manual Investigation
- after the decommission actor run we cancel the loop if decommission is successful. In case of an error we do not cancel the loop.  
- after fixing the issue, the decommission gets stuck in an error, but the node is not terminating.
- running exec on a pod  with command:
```
bash-4.4$ ./cockroach node decommission 4 --wait=none --certs-dir=cockroach-certs
ERROR: server closed the connection.
Is this a CockroachDB node?
failed to connect to the node: initial connection heartbeat failed: rpc error: code = Unavailable desc = connection error: desc = "transport: authentication handshake failed: context deadline exceeded"
Failed running "node decommission"
bash-4.4$
```
- running another command works
```
bash-4.4$ ./cockroach node status --certs-dir=cockroach-certs
  id |                 address                 |               sql_address               |  build  |         started_at         |         updated_at         | locality | is_available | is_live
-----+-----------------------------------------+-----------------------------------------+---------+----------------------------+----------------------------+----------+--------------+----------
   1 | cockroachdb-0.cockroachdb.default:26258 | cockroachdb-0.cockroachdb.default:26257 | v21.1.1 | 2021-06-02 16:32:42.104195 | 2021-06-02 16:56:01.625377 |          | true         | true
   2 | cockroachdb-2.cockroachdb.default:26258 | cockroachdb-2.cockroachdb.default:26257 | v21.1.1 | 2021-06-02 16:32:43.685864 | 2021-06-02 16:55:58.700637 |          | true         | true
   3 | cockroachdb-1.cockroachdb.default:26258 | cockroachdb-1.cockroachdb.default:26257 | v21.1.1 | 2021-06-02 16:32:43.786401 | 2021-06-02 16:55:58.798963 |          | true         | true
   4 | cockroachdb-3.cockroachdb.default:26258 | cockroachdb-3.cockroachdb.default:26257 | v21.1.1 | 2021-06-02 16:32:44.472166 | 2021-06-02 16:55:59.48676  |          | true         | true
(4 rows)
bash-4.4$ 
```
### Conclusion
- the CLI connects to the GRPC port - when we moved from SQL and GRPC sharing 26257 to having GRPC running on 26258, we need to actually explicitly declare the port to connect to on the CLI.


keith-mcclellan (changes requested):
I'm not sure this is quite right... a node thats currently draining is going to look like:
```
 id | is_live | replicas | is_decommissioning | is_draining  
+---+---------+----------+--------------------+-------------+
  4 |  true   |       45 |        true        |    false     
  ```
  
  and a node that's completely decommissioned and ready to be removed from the cluster will look like:
  
  ```
   id | is_live | replicas | is_decommissioning | is_draining  
+---+---------+----------+--------------------+-------------+
  4 |  true   |        0 |        true        |    false    
  ```
  
  So there's a valid state where a node is both live and decommissioning, where the replica count > 0.  This just means the decommissioning is still ongoing.
  
  The invalid state is where decommissioning is true, replica count is > 0, and is_live is FALSE.  That means the node died during decommission and the decommission failed

**Comment by keith-mcclellan** [comment]
see https://www.cockroachlabs.com/docs/v21.1/remove-nodes.html

File: pkg/testutil/require.go [hunk 1]
```diff
@@ -293,11 +296,90 @@ func RequireDecommissionNode(t *testing.T, sb testenv.DiffingSandbox, b ClusterB
...
+		if err != nil {
+			t.Logf("makeDrainStatusChecker failed due to error %v\n", err)
+			return false, err
+		}
 		return true, nil
 	})
 	require.NoError(t, err)
 }
 
+func makeDrainStatusChecker(t *testing.T, sb testenv.DiffingSandbox, b ClusterBuilder, numNodes uint64) error {
```
chrislovecnm, line 323: This is great and should be part of our code that we check this upon deco

chrislovecnm (approved):
Is this passing with the new fix?

pkg/testutil/require.go
```diff
@@ -293,11 +296,90 @@ func RequireDecommissionNode(t *testing.T, sb testenv.DiffingSandbox, b ClusterB
...
+
+	stdout, stderror, err := kube.ExecInPod(sb.Mgr.GetScheme(), sb.Mgr.GetConfig(), cluster.Namespace(),
+		fmt.Sprintf("%s-0", cluster.StatefulSetName()), resource.DbContainerName, cmd)
+	if err != nil {
+		t.Logf("exec cmd = %s on pod= %s-0 exit with error %v and stdError %s", cmd, cluster.StatefulSetName(), err, stderror)
+		return err
+	}
+	if stderror != "" {
+		t.Logf("exec cmd = %s on pod= %s-0 exit with error %v and stdError %s", cmd, cluster.StatefulSetName(), err, stderror)
+		return err
```
udnay, line 335: This will always be `nil` maybe we need to create a new error specific to this?

pkg/testutil/require.go
```diff
@@ -293,11 +296,90 @@ func RequireDecommissionNode(t *testing.T, sb testenv.DiffingSandbox, b ClusterB
...
+		if err != nil {
+			return errors.Wrapf(err, "failed to get node draining status")
+		}
+
+		idStr, isLive, replicasStr, isDecommissioning := record[0], record[8], record[9], record[10]
+		id, err := strconv.ParseUint(idStr, 10, 32)
+		if err != nil {
+			return errors.Wrap(err, "failed to extract node id from string")
+		}
+		if id <= numNodes {
```
udnay, line 357: So the logic is here that we always decommission the node with the highest ID first? This is a bit of magic to me, but maybe its fine. For a later PR but maybe we add a check in here to make sure no other nodes are decommissioning?

pkg/testutil/require.go
alinadonisa: stderror is a string and contains the stderror content from the command we exec in bash. In case of a failure of the bash command we can debug better.

keith-mcclellan (approved):
LGTM

pkg/testutil/require.go
udnay: Right, but just above we check `err != nil` and return the error, so the `stderror` check is only reached if `err == nil` and then we return `nil` so from the calling functions perspective the function exited successfully even though we exited out early.

PR #572, opened 2021-06-10
Removing pvc arguments in Makefile

chrislovecnm:
Missed removing pvc test argument when I removed the argument from the
tests themselves.


udnay (approved):
LGTM

PR #579, opened 2021-06-11
fix decommision e2e test to use host

alinadonisa:
In the e2e tests for decommission we used the index of the pod to check the replicas of the decommissioned nodes. 
Since we are using Pod Parallel Strategy the index of the pod and the  id are not the same anymore, so now I am filtering using the address that contains the host of the last node that was decommissioned to get the replicas.


keith-mcclellan (approved):
LGTM

PR #585, opened 2021-06-15
Unit test for testing actor order

chrislovecnm:
The slice that is returned by NewOperatorActions drives
the order of actors working on a CRD. This order
is INCREDIBLY important for the operator, so this test
ensures the order is not changed accidentally.

### DIFF
--- a/pkg/actor/actor_test.go
+++ b/pkg/actor/actor_test.go
@@ -1,0 +1,59 @@
+/*
+Copyright 2021 The Cockroach Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


pkg/actor/actor_test.go
```diff
@@ -0,0 +1,59 @@
...
+	"reflect"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+)
+
+func TestActorOrder(t *testing.T) {
+
+	// So the actor order is so important that we need to ensure that the slice
```
udnay, line 29: I'd use caps :)

udnay (approved):
LGTM

PR #595, opened 2021-06-16
Was not checking that client custom cert was set

chrislovecnm:
We were checking that the node custom cert was set and not
the client custom cert. We should have still returned the
client custom cert, but this still is a bug.

I filed and issue for checking that both API custom cert
feilds are set, which we should probably do with an
addmission controller.


keith-mcclellan (approved):
LGTM

PR #597, opened 2021-06-16
Retrying if we have a conflict with an update

chrislovecnm:
The intial update of the sts returned that we cannot update the
sts since the sts is newer.  We then get the sts again and retry to
update it again.


keith-mcclellan (approved):
LGTM

PR #602, opened 2021-06-17
Enabling Decommission

chrislovecnm:
- enabled deco via changing feature gate
- changed unit test, to test that it is enabled
- updated operator_features comments
- moved prune pvc to alpha
- we have two e2e tests, one that prunes PVCs and another that
  does not prune pvcs.

### DIFF
--- a/e2e/decomission/decomission_test.go
+++ b/e2e/decomission/decomission_test.go
@@ -49,6 +49,59 @@ func TestMain(m *testing.M) {
 	os.Exit(code)
 }
 
+// TODO once prune pvc feature gate is set to "true" by default, we can
+// remove this test.
+
+// TestDecomissionFunctionalityWithPrune creates a cluster of 4 nodes and then decomissions on of the CRDB nodes.
+// It then checks that the cluster is stable and that decomissioning is successful.
+func TestDecommissionFunctionalityWithPrune(t *testing.T) {
+
+	// Testing removing and decommisioning a node.  We start at 4 node and then
+	// remove the 4th node
+
+	// turn on featuregate since Decommission is disabled by default currently
+	utilfeature.DefaultMutableFeatureGate.Set("AutoPrunePVC=true")
+
+	if testing.Short() {
+		t.Skip("skipping test in short mode.")
+	}
+	// Does not seem to like running in parallel
+	if parallel {
+		t.Parallel()
+	}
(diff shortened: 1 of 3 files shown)


e2e/decomission/decomission_test.go
```diff
@@ -49,15 +49,68 @@ func TestMain(m *testing.M) {
...
 // It then checks that the cluster is stable and that decomissioning is successful.
 func TestDecommissionFunctionality(t *testing.T) {
 
 	// Testing removing and decommisioning a node.  We start at 4 node and then
 	// remove the 4th node
 
-	// turn on featuregate since Decommission is disabled by default currently
-	utilfeature.DefaultMutableFeatureGate.Set("UseDecommission=true,AutoPrunePVC=true")
+	// making sure the feature gate is off for prunePVC
+	utilfeature.DefaultMutableFeatureGate.Set("AutoPrunePVC=false")
```
udnay, line 113: On this test can we check to make sure there is an orphaned PV?

e2e/decomission/decomission_test.go
```diff
@@ -49,15 +49,68 @@ func TestMain(m *testing.M) {
...
+			Test: func(t *testing.T) {
+				current := builder.Cr()
+				require.NoError(t, sb.Get(current))
+
+				current.Spec.Nodes = 3
+				require.NoError(t, sb.Update(current))
+				testutil.RequireClusterToBeReadyEventuallyTimeout(t, sb, builder, 500*time.Second)
+				testutil.RequireDecommissionNode(t, sb, builder, 3)
+				testutil.RequireDatabaseToFunction(t, sb, builder)
+				t.Log("Done with decommision")
```
udnay, line 98: The opposite here, that no PVC remains

chrislovecnm: What am I missing?

udnay: I think the logic doesn't immediately prune the PVC so we probably can't check for this anyhow. What I was thinking of was something like `assert.Equal(t, numPVCsForCRDB, 3)` or something. The `RequireDecommissionNode` checks CRDB to validate the operation is completed but doesn't check the orphaned PVC

udnay: I don't think it's testable here tbh.

udnay (approved):
LGTM

PR #607, opened 2021-06-18
Fixing nil on Job selector

chrislovecnm:
We are getting a nil when we get the Job from the API.  This code
runs a list, which reconciles the API, and gets the selector.


udnay (approved):
LGTM

PR #608, opened 2021-06-18
Adding logging and cleaning up code

chrislovecnm:
Added another check and added some logging.

### DIFF
--- a/pkg/actor/validate_version.go
+++ b/pkg/actor/validate_version.go
@@ -150,14 +150,25 @@ func (v *versionChecker) Act(ctx context.Context, cluster *resource.Cluster) err
 			return err
 		}
 	}
-	if job == nil || job.Spec.Selector == nil {
+
+	// We have hit an edge case were sometimes the job selector is nil, the following block is extra code
+	// that tries to list the jobs and then get the job again, which probably will reconcile
+	// the API.  We also removed setting the job selector as well.
+	if job.Spec.Selector == nil {
+		log.V(int(zapcore.DebugLevel)).Info("Job or Job Selector returned as nil, attempting to get it again.")
 
 		// The job is nil or the selector is nil, we are doing a list, which
 		// should reconcile the API and we we do another get the job should have
 		// the selector.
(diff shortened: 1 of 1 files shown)


pkg/actor/validate_version.go
```diff
@@ -150,27 +150,42 @@ func (v *versionChecker) Act(ctx context.Context, cluster *resource.Cluster) err
 			return err
 		}
 	}
+
+	// We have hit an edge case were sometimes the job selector is nil, the following block is extra code
+	// that tries to list the jobs and then get the job again, which probably will reconcile
+	// the API.  We also removed setting the job selector as well.
 	if job == nil || job.Spec.Selector == nil {
```
udnay, line 157: `job` can never be `nil` we assign it on line line 139, or am I missing something?

pkg/actor/validate_version.go
```diff
@@ -150,27 +150,42 @@ func (v *versionChecker) Act(ctx context.Context, cluster *resource.Cluster) err
...
+			return err
+		}
+
 		if err := v.client.Get(ctx, key, job); err != nil {
 			log.Error(err, "unable to get job")
 			return err
 		}
 	}
 
 	if job == nil {
```
udnay, line 181: Same comment as above

pkg/actor/validate_version.go
chrislovecnm: The “Get” could assign it to nil

udnay: How does that work? In the `Get` we're passing in a pointer so that the `Get` can modify fields inside of `job` but inside of `Get` I can't assign `obj = nil` and then have `job` in calling scope become `nil`, what am I missing? Is `obj` in `Get` actually a `**Job`? In the comments it just says pointer `obj must be a struct pointer`

chrislovecnm: Changed :)

pkg/actor/validate_version.go
chrislovecnm: changed

udnay (approved):
LGTM

PR #618, opened 2021-06-30
Updating operate manifest with PVC prune feature gate

chrislovecnm:


udnay:
@taroface this is part of the documentation right? We should merge this for GA in that case.

keith-mcclellan (approved):
LGTM

Issue #619, opened 2021-06-30
auto-pruning is not working as intended


taroface:
Blocks #598.

On GKE, using the current published [Operator](https://github.com/cockroachdb/cockroach-operator/blob/master/manifests/operator.yaml), scaling down a cluster from 4 to 3 nodes also removes both the PVC and PV bound to the pod that is removed.

`kubectl get events` shows the following error referencing the PV that was removed:

```
74s         Normal    VolumeDelete              persistentvolume/pvc-63ce836a-1258-4c58-8b37-d966ed12d50a   googleapi: Error 400: The disk resource 'projects/rk-k8s-test/zones/us-east4-a/disks/gke-cockroachdb-13be10-pvc-63ce836a-1258-4c58-8b37-d966ed12d50a' is already being used by 'projects/rk-k8s-test/zones/us-east4-a/instances/gke-cockroachdb-default-pool-8532dad5-m9k2', resourceInUseByAnotherResource
```

@udnay suggests this might have to do with the reclaim policy being set to Delete.

chrislovecnm:
@taroface this should be addressed. We can test when we get another release

udnay:
Fixed by #620

PR #620, opened 2021-07-01
Add missing feature flag guard

udnay:
This is a fix for #619 when the initial feature flag went in we guarded only against the initial PVC prune and not the final prune.


chrislovecnm (approved):
LGTM

PR #625, opened 2021-07-08
Add pvc check to decommission

udnay:
Added a check to make sure the PVC's are deleted or kept as expected. Also changed the decommission test cases to use the namespace. It was causing some requests like for PVCs to select across all namespaces and was finding extras. [cluster.Namespace()](https://github.com/cockroachdb/cockroach-operator/blob/master/pkg/testutil/require.go#L516) was always `nil`. We might want to check if this is causing issues in other testacases as well.

### DIFF
--- a/pkg/testutil/require.go
+++ b/pkg/testutil/require.go
@@ -598,3 +598,32 @@ func getPodLog(ctx context.Context, podName string, namespace string, clientset 
 	}
 	return buf.String(), nil
 }
+
+// RequirePVCToResize checks that the PVCs are resized correctly
+func RequirePVCNumbers(t *testing.T, sb testenv.DiffingSandbox, b ClusterBuilder, quantity int32) {
+	cluster := b.Cluster()
+
+	// TODO rewrite this
+	err := wait.Poll(10*time.Second, 500*time.Second, func() (bool, error) {
+		clientset, err := kubernetes.NewForConfig(sb.Mgr.GetConfig())
+		require.NoError(t, err)
+
+		sts, err := fetchStatefulSet(sb, cluster.StatefulSetName())
+
+		selector, err := metav1.LabelSelectorAsSelector(sts.Spec.Selector)
+		if err != nil {
+			return false, err
+		}
+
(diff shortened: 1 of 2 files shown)


pkg/testutil/require.go
```diff
@@ -598,3 +598,41 @@ func getPodLog(ctx context.Context, podName string, namespace string, clientset
 	}
 	return buf.String(), nil
 }
+
+// RequirePVCToResize checks that the PVCs are resized correctly
+func RequireNumberOfPVCs(t *testing.T, ctx context.Context, sb testenv.DiffingSandbox, b ClusterBuilder, quantity int) {
+	cluster := b.Cluster()
+	var boundPVCCount = 0
+
+	// TODO rewrite this
```
chrislovecnm, line 607: Need more details, and also an issue please

pkg/testutil/require.go
```diff
@@ -598,3 +598,41 @@ func getPodLog(ctx context.Context, podName string, namespace string, clientset
...
+			LabelSelector: selector.String(),
+		})
+
+		for _, pvc := range pvcs.Items {
+			if pvc.Status.Phase == corev1.ClaimBound {
+				boundPVCCount = boundPVCCount + 1
+			}
+		}
+
+		if err != nil {
```
chrislovecnm, line 629: Move this to line 618, also log the error on both of the checks please

chrislovecnm: Unless to require catches it

chrislovecnm (commented):
Some changes that I noted

pkg/testutil/require.go
udnay: This is copy paste from the esize PVCs but i'll add an issue to cover both

udnay: Created #631

pkg/testutil/require.go
udnay: The require below on 635 will catch the error and print it.

Moved the `if` up, nice catch.

chrislovecnm (approved):
LGTM

PR #626, opened 2021-07-12
Adding Affinity Rules to API

chrislovecnm:
- added the new API value "affinity"
- affinity is a corev1.Affinity struct which allows the user to set
pod scheduling rules
- new unit test
- created a new feature gate AffinityRules which is now disabled


pkg/resource/testdata/TestStatefulSetBuilder/insecure_statefulset_cli_args_in.yaml
```diff
@@ -40,4 +40,16 @@ spec:
   topology:
     zones:
       - locality: ""
+  affinity:
```
chrislovecnm, line 43: This is not working .. arg

config/crd/bases/crdb.cockroachlabs.com_crdbclusters.yaml
```diff
@@ -55,6 +55,347 @@ spec:
                 items:
                   type: string
                 type: array
+              affinity:
```
udnay, line 58: Is this the same schema as for pod affinity rules? Can we just link to that? If we're just passing it through is there anyway to specify that in the yaml and explicitly track that API?

chrislovecnm: Don’t quite get what you are saying

pkg/resource/statefulset_test.go
```diff
@@ -38,6 +39,8 @@ import (
 var update = flag.Bool("update", false, "update the golden files of this test")
 
 func TestStatefulSetBuilder(t *testing.T) {
+	// turn on featuregate to test rules
+	utilfeature.DefaultMutableFeatureGate.Set("AffinityRules=true")
```
udnay, line 43: Is there a test where we have the affinity rules set in a cluster spec but the feature flag is disabled and things work as expected? i.e. the affinity rules are not passed to the pod

udnay (approved):
A couple questions but otherwise LGTM

config/crd/bases/crdb.cockroachlabs.com_crdbclusters.yaml
udnay: I don't think it's possible but just saying
```
affinity: *podSpecAffinitySchema
```
Linking to that block of yaml rather than copy pasting the whole thing

chrislovecnm: We are using the strict from K8s in the api

pkg/resource/statefulset_test.go
chrislovecnm: We can test that the feature is disabled, but again we have test for the featuregates already

PR #627, opened 2021-07-12
Do not override `CHANNEL` and `IS_DEFAULT_CHANNEL`

rail:
Previously, we hardcoded the `CHANNEL` and `IS_DEFAULT_CHANNEL` variables
in Makefile for the `release/gen-files` target in order to reduce the
possibility of publishing a release to a wrong channel.

Now, that we have these steps documented and automated, we can remove
the hardcoded variables. By default, they will be set to `CHANNEL=beta`
and `IS_DEFAULT_CHANNEL=0`. The release docs will explicitly override
these variables.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -239,8 +239,8 @@ release/gen-templates:
 # operator version is changed. The results are committed to Git.
 .PHONY: release/gen-files
 release/gen-files: release/gen-templates
-	$(MAKE) CHANNEL=beta IS_DEFAULT_CHANNEL=0 release/update-pkg-manifest && \
-	$(MAKE) CHANNEL=beta IS_DEFAULT_CHANNEL=0 release/opm-build-bundle && \
+	$(MAKE) release/update-pkg-manifest && \
+	$(MAKE) release/opm-build-bundle && \
 	git add . && \
 	git commit -m "Bump version to $(VERSION)"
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -239,8 +239,8 @@ release/gen-templates:
 # operator version is changed. The results are committed to Git.
 .PHONY: release/gen-files
 release/gen-files: release/gen-templates
-	$(MAKE) CHANNEL=beta IS_DEFAULT_CHANNEL=0 release/update-pkg-manifest && \
-	$(MAKE) CHANNEL=beta IS_DEFAULT_CHANNEL=0 release/opm-build-bundle && \
+	$(MAKE) release/update-pkg-manifest && \
```
chrislovecnm, line 242: What variables are we using now?

rail: I will update the release doc to use `make releaase/gen-files CHANNEL=stable IS_DEFAULT_CHANNEL=1` for the 2.0.x release.

rail: Looks like I need to export these variables, because we recursively call `$(MAKE)`.

rail: Actually it works fine even without exporting these variables. All good.

keith-mcclellan (approved):
LGTM

PR #628, opened 2021-07-12
Set maturity to stable

rail:


keith-mcclellan (approved):
LGTM

udnay:
LGTM, do we flip this back to beta once we start doing more feature work or are we doing a 2.0.0 release branch?

rail:
> LGTM, do we flip this back to beta once we start doing more feature work or are we doing a 2.0.0 release branch?

Not sure... Maybe we should make this overridable?

udnay:
I think it's fine for today so that we don't add in too much complexity. I'd like to err on the side of simplicity for trying to publish this to openshift today. If we do need to make it overridable then lets create a ticket for it.

PR #634, opened 2021-07-13
Bump version to 2.0.1

rail:


chrislovecnm:
Do we need to run openshift CI on this?

rail:
> Do we need to run openshift CI on this?

This is a good idea. We can force a nightly run via https://teamcity.cockroachdb.com/buildConfiguration/Cockroach_CockroachOperator_Nightlies_GkeNightly?mode=builds

keith-mcclellan:
We missed updating latest so I just did that.  This should be ready to submit once the red hat outage is over

keith-mcclellan (approved):
LGTM

PR #641, opened 2021-07-20
Removing initcontainer when TLS is disabled

chrislovecnm:
Fixing bug where the operator was adding the initcontainer
when TLS was disabled.


keith-mcclellan (approved):
LGTM

PR #652, opened 2021-07-26
Adding Downgrade Test

chrislovecnm:
- adding a new e2e test that upgrades a cluster and then
downgrades the same cluster


keith-mcclellan (commented):
Was the intent here to just do the patch upgrade/downgrade? Are we going to add a separate test for feature release downgrades or should we do that now?

What you did LGTM but its only for patches, feature releases are more complex

keith-mcclellan (approved):
LGTM, we opened a separate PR for the feature release downgrade unit test

PR #653, opened 2021-07-27
Adding unit tests for checking preserve downgrade

chrislovecnm:
- refactored code so that we can write the unit tests without using a db
- also created a public func that we can use during e2e testing to
ensure that downgrade was actually set
- created a unit test to test the perserve option setting
- this work does not check that the perserve was set


keith-mcclellan:
@chrisseto  Can I get your input on this one with @udnay being out please?

pkg/update/update_cockroach_version.go
```diff
@@ -255,6 +240,30 @@ func kindAndCheckPreserveDowngradeSetting(
 	return "UNKNOWN", err
 }
 
+// CheckDowngradeSetting retrieves the downgrade setting from the database and then tests that the update is feasible.
+// This func will return an error if the update is not to the Major and Minor version that is requested.
+func CheckDowngradeSetting(ctx context.Context, wantVersion *semver.Version, currentVersion *semver.Version, db *sql.DB) (string, error) {
+	s := "MAJOR_ROLLBACK"
```
chrislovecnm, line 246: I need to factor this to a const

pkg/update/update_cockroach_version_test.go
```diff
@@ -0,0 +1,74 @@
...
+package update
+
+import (
+	"testing"
+
+	semver "github.com/Masterminds/semver/v3"
+	"github.com/stretchr/testify/require"
+)
+
+func TestPerserveMatches(t *testing.T) {
```
chrislovecnm, line 26: I need to fix the typos

pkg/update/update_cockroach_version_test.go
```diff
@@ -0,0 +1,74 @@
...
+		wantVersion     *semver.Version
+		currentVersion  *semver.Version
+		perserveVersion *semver.Version
+		result          bool
+	}{
+		{
+			"perserved is not set correctly",
+			semver.MustParse("20.1.6"),
+			semver.MustParse("21.1.5"),
+			semver.MustParse("20.2.5"),
```
chrisseto, line 38: Are these tests correct? My understanding is that preserve downgrade versions will only be the major.minor IE 20.1, 21.2, etc. Will this logic or something further upstream break if a patch version doesn't exist?

https://www.cockroachlabs.com/docs/v21.1/upgrade-cockroach-version.html#step-3-decide-how-the-upgrade-will-be-finalized

pkg/update/update_cockroach_version_test.go
chrisseto: nit: `TestDowngradeAllowed` would be a bit more clear.

pkg/update/update_cockroach_version_test.go
```diff
@@ -0,0 +1,74 @@
...
+	semver "github.com/Masterminds/semver/v3"
+	"github.com/stretchr/testify/require"
+)
+
+func TestPerserveMatches(t *testing.T) {
+	tests := []struct {
+		description     string
+		wantVersion     *semver.Version
+		currentVersion  *semver.Version
+		perserveVersion *semver.Version
```
chrisseto, line 31: nit: `preserveVersion`

chrisseto (approved):
LGTM provided you make the changes you yourself have called out.

PR #654, opened 2021-07-27
Using code to check that preserve is set

chrislovecnm:
- now using code that checks the preserve is set in the database
  when doing a major upgrade

### DIFF
--- a/pkg/kube/helpers.go
+++ b/pkg/kube/helpers.go
@@ -215,6 +215,24 @@ func mutate(f MutateFn, key client.ObjectKey, obj client.Object) error {
 // TODO this code is from https://github.com/kubernetes/kubernetes/blob/master/pkg/api/v1/pod/util.go
 // We need to determine if this functionality is available via the client-go
 
+// IsPodAvailable returns true if a pod is available; false otherwise.
+// Precondition for an available pod is that it must be ready. On top
+// of that, there are two cases when a pod can be considered available:
+// 1. minReadySeconds == 0, or
+// 2. LastTransitionTime (is set) + minReadySeconds < current time
+func IsPodAvailable(pod *corev1.Pod, minReadySeconds int32, now metav1.Time) bool {
+	if !IsPodReady(pod) {
+		return false
+	}
+
+	c := GetPodReadyCondition(pod.Status)
+	minReadySecondsDuration := time.Duration(minReadySeconds) * time.Second
+	if minReadySeconds == 0 || (!c.LastTransitionTime.IsZero() && c.LastTransitionTime.Add(minReadySecondsDuration).Before(now.Time)) {
+		return true
+	}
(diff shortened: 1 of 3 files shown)


pkg/kube/helpers.go
```diff
@@ -215,6 +215,24 @@ func mutate(f MutateFn, key client.ObjectKey, obj client.Object) error {
 // TODO this code is from https://github.com/kubernetes/kubernetes/blob/master/pkg/api/v1/pod/util.go
 // We need to determine if this functionality is available via the client-go
 
+// IsPodAvailable returns true if a pod is available; false otherwise.
+// Precondition for an available pod is that it must be ready. On top
+// of that, there are two cases when a pod can be considered available:
+// 1. minReadySeconds == 0, or
+// 2. LastTransitionTime (is set) + minReadySeconds < current time
+func IsPodAvailable(pod *corev1.Pod, minReadySeconds int32, now metav1.Time) bool {
```
chrislovecnm, line 223: @chrisseto this seems to still fail.  I am getting db connection issues, and the only thing I can think of is that the pod is not up and running correctly.  Thoughts?

pkg/testutil/require.go
```diff
@@ -228,23 +235,57 @@ func testPodsWithPredicate(pods []corev1.Pod, pred func(*corev1.Pod) bool) bool
 	return true
 }
 
-func statefulSetIsReady(ss *appsv1.StatefulSet) bool {
-	return ss.Status.ReadyReplicas == ss.Status.Replicas
-}
+func statefulSetIsReady(t *testing.T, sts *appsv1.StatefulSet, sb testenv.DiffingSandbox) (bool, error) {
+	if sts.Status.ReadyReplicas != sts.Status.Replicas {
```
chrisseto, line 239: I think you might want an additional check of `.Generation > .Status.ObservedGeneration` here? It might be possible that you're picking up old pods?

pkg/testutil/require.go
```diff
@@ -228,23 +235,57 @@ func testPodsWithPredicate(pods []corev1.Pod, pred func(*corev1.Pod) bool) bool
...
+	}
+	clientset, err := kubernetes.NewForConfig(sb.Mgr.GetConfig())
+	if err != nil {
+		t.Log("error building k8s api config")
+		return false, err
+	}
 
-// TODO we are not using this
+	// Get all pods
+	podList, err := clientset.CoreV1().Pods(sts.Namespace).List(context.TODO(), metav1.ListOptions{})
```
chrisseto, line 250: Wouldn't you want to limit this to pods owned by this statefulset?

pkg/kube/helpers.go
chrisseto: Seems like you might be grabbing the wrong pods?

PR #660, opened 2021-08-04
remove labels on client pod

### LINKED ISSUE #657: Examples/client-secure-operator.yaml breaks service on example.yaml
Currently the stable documentation shows how to set up using example.yaml, and then access it with client-secure-operator.yaml.
The labels applied on client-secure-operator seems to be identical to what the operator creates for the cockroachdb-public service, so requests will occasionally end up loadbalanced onto the client pod, causing spurious "connection refused" errors that were very hard to debug.

I have not yet checked if this is fixed if you rename the instance in example to something else than cockroachdb, but is there a reason for the labels applied to the client at all?

taroface:
Fixes #657.

### DIFF
--- a/examples/client-secure-operator.yaml
+++ b/examples/client-secure-operator.yaml
@@ -16,10 +16,6 @@ apiVersion: v1
 kind: Pod
 metadata:
   name: cockroachdb-client-secure
-  labels:
-    app.kubernetes.io/component: database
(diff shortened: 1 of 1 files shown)


chrislovecnm:
@taroface can you sign the CLA?

chrislovecnm (approved):
LGTM

PR #664, opened 2021-08-06
Add GitHub action to update CRDB versions

rail:
Previously, we had to manually track CRDB releases and update the
corresponding file in this repo.

This patch uses the RedHat Catalog API to fetch all published CRDB
images and regenerate the `crdb-versions.yaml` file. If the file is
changed, the GitHub action would create a PR.

Example PR: https://github.com/rail/cockroach-operator/pull/3

### DIFF
--- a/hack/update-crdb-versions.sh
+++ b/hack/update-crdb-versions.sh
@@ -1,0 +1,54 @@
+#!/usr/bin/env bash
+
+# Copyright 2021 The Cockroach Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


rail:
Next step would be creating a release PR.

hack/update-crdb-versions.sh
```diff
@@ -0,0 +1,54 @@
...
+# This file contains a list of CockroachDB versions that are supported by the
+# operator. hack/crdbversions/main.go uses this list to generate various
+# manifests.
+# Please update this file when CockroachDB releases new versions.
+
+CrdbVersions:
+EOF
+
+# Skip unsupported versions and the latest tag
+for version in $(curl $URL | jq -r '.data[] .repositories[] .tags[] .name' | grep -v ^v19 | grep -v latest | grep -v ubi$ | sort --version-sort); do
```
udnay, line 52: /nit Can the `^v19` be a parameter? maybe 
`EXCLUDE_VERSION={$1:-"v19"}` or something?

udnay (approved):
Other than nit LGTM

hack/update-crdb-versions.sh
rail: I'd say let's leave it as is for now. I'm going to revise this script in a few weeks after letting it prove itself. Next iteration will be creating a release PR whenever we detect a new CRDB version.

PR #669, opened 2021-08-17
Run bazel run //:gazelle to update BUILD files

pseudomuto:


pseudomuto:
nvm, I found `//hack:update-bazel`

PR #676, opened 2021-08-18
Add kuberecord to enable VCR for K8s API calls

pseudomuto:
__**NOTE**: I know +1600 looks scary, but over 900 lines are for the YAML fixture__

Adding `kuberecord` which is a package that leverages the [go-vcr](https://github.com/dnaeon/go-vcr) to record interactions with a K8s API server.

This is useful to avoid having to run every command over and over again when running tests. In theory, unless the API version changes, we typically don't expect the behaviour to be any different for things like CreateNamespace, ListPods, etc.

Kuberecord works by wrapping the incoming rest.Config to handle loading prerecorded interactions, scrubbing sensitive/flakey headers, and storing results to disk (`./testdata/<name>.vcr` by default).

This code is largely "borrowed" from an internal repo where it has been in use for a while already. In order to keep the diff here more reasonable, I haven't updated any tests to use it just yet. I'm happy to start addressing that in future PRs once we're happy with the implementation here.

See _record_test.go_ in this PR for an example of how to use it.

### DIFF
--- a/pkg/kuberecord/cassettes_test.go
+++ b/pkg/kuberecord/cassettes_test.go
@@ -1,0 +1,114 @@
+/*
+Copyright 2021 The Cockroach Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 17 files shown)


pkg/kuberecord/cassettes_test.go
```diff
@@ -0,0 +1,99 @@
...
+			URL:     url,
+			Headers: http.Header{fpHeader: []string{fp}},
+		}
+	}
+
+	tests := []struct {
+		name     string
+		req      *http.Request
+		cassette cassette.Request
+		fail     bool
```
udnay, line 64: Can this not be `pass` so we don't need the negation?

udnay (commented):
Can you add a sample test on something like rolling restart or another action to give a good example.

udnay (approved):
LGTM

PR #680, opened 2021-08-19
Run `go mod tidy`

davidwding:
Housekeeping: run `go mod tidy` on a clean branch. This will make my other PR cleaner.


chrislovecnm:
@davidwding any ideas why  e2e-kind-decomission failed?

davidwding:
> @davidwding any ideas why e2e-kind-decomission failed?

I don't...it seems to have gotten to the step where it decommissions a node and then just got stuck for an hour before timing out. Lemme investigate some more...

davidwding:
It passed after rerunning upon merging in master. Have you observed any flakiness with the test before?

chrislovecnm:
> It passed after rerunning upon merging in master. Have you observed any flakiness with the test before?

Yes, we have had flakes, but not for a while.  Should you grab the logs and open an issue so that we don't lose looking at the flake?  The problem as well is that GH Actions with testings are very small instances.  We sometimes resource starve the cluster, so that we get problems.  Also, in the past, there were times where GH testing was not stable and having a bad hair day.  Over the last couple of months, testing has been quite solid and not flakey.

davidwding:
> > It passed after rerunning upon merging in master. Have you observed any flakiness with the test before?
> 
> Yes, we have had flakes, but not for a while. Should you grab the logs and open an issue so that we don't lose looking at the flake? The problem as well is that GH Actions with testings are very small instances. We sometimes resource starve the cluster, so that we get problems. Also, in the past, there were times where GH testing was not stable and having a bad hair day. Over the last couple of months, testing has been quite solid and not flakey.

I've grabbed the logs and opened a bug https://github.com/cockroachdb/cockroach-operator/issues/685

Any concerns about this cleanup PR otherwise?

pseudomuto (approved):

PR #681, opened 2021-08-19
Remove unused code

pseudomuto:
Removing some unused code from `pkg/...`. I found these by using `staticcheck -check U1000 ./pkg/...`.


davidwding (dismissed):
approving because i love removing dead code, but also defer to chris on whether any of this is actually worth preserving still

udnay (approved):
LGTM 🔥

PR #682, opened 2021-08-20
Slight refactor of part of e2e pvcresize test

### LINKED ISSUE #631: Rewrite polling logic for fetching stateful set in e2e tests
The polling loop used in [RequirePVCToResize](https://github.com/cockroachdb/cockroach-operator/blob/0b67f1f87db452de1d1e3acf72f11e9e55182610/pkg/testutil/require.go#L471) and [RequireNumberOfPVCs](https://github.com/cockroachdb/cockroach-operator/blob/0b67f1f87db452de1d1e3acf72f11e9e55182610/pkg/testutil/require.go#L608) should be rewritten. 

It might make sense to pull all of this logic out into a helper to fetch the required k8s objects for the require tests and get rid of the current loops. Separate the fetching of the objects from the checks themselves.

davidwding:
This is a follow up to https://github.com/cockroachdb/cockroach-operator/pull/677.

The PVC check within the pvcresize e2e test has been slightly refactored to untangle fetching the PVCs and running the test logic itself. Tested using `make test/e2e/gke`.

Fixes #631.


pkg/testutil/require.go
```diff
@@ -421,74 +421,56 @@ func getCount(t *testing.T, rows *sql.Rows) (count int) {
...
+	var pvcsToKeep map[string]bool
 
-	// TODO rewrite this
 	err := wait.Poll(10*time.Second, 500*time.Second, func() (bool, error) {
 		ss, err := fetchStatefulSet(sb, cluster.StatefulSetName())
 		if err != nil {
 			return false, err
 		}
-
 		if ss == nil {
```
pseudomuto, line 450: /nit looking at `fetchStatefulSet` I don't think this'll ever be `nil`

pseudomuto (approved):
Just the nit about an unnecessary `nil` check, otherwise LGTM 👍

pseudomuto (approved):

PR #683, opened 2021-08-23
Update env.Env to be universal

pseudomuto:
For most things, we're using `env.Env` as the test environment, which performs a `paths.MaybeSetPath` using kubetest2-kind as the lookup variable. This will check the _PATH_ to see if the parent directory is included and add it if it's missing.

**The problem:**

The api tests (for example) use the standard envtest provided by controller-runtime to create an environment. As we beef up the env in this project, it'd be nice to leverage the same env for all tests.

Also, it seems that all e2e tests and the api ones require that a few executables be available, more specifically things like kubectl, kubeapi-server, and kind. Rather than looking for `kubetest2-kind` and doing a `MaybeSetEnv` or individually setting `TEST_KUBECTL`, etc. I think we could leverage `KUBEBUILDER_ASSETS` and `PATH`.

**The solution:**

Drop the _testutil/paths_ package and add `env.ExpandPath` as well as `env.PrependPath`. These are then used to set `KUBEBUILDER_ASSETS` to the _hack/bin_ directory in the test workspace (also works without bazel) and to ensure that directory is on the PATH.

This path is defined by the `binaries` flag, which can be overridden if necessary using the `args` key in a `go_test` target.


pkg/testutil/env/path.go
```diff
@@ -0,0 +1,51 @@
...
+		prefix := []string{v, os.Getenv("TEST_WORKSPACE"), os.Getenv("BUILD_WORKSPACE_DIRECTORY")}
+		return filepath.Join(append(prefix, path...)...)
+	}
+
+	// when not running with bazel
+	return filepath.Join(append([]string{ProjectDir()}, path...)...)
+}
+
+// PrependPath prepends the supplied path to PATH
+func PrependPath(path ...string) {
```
davidwding, line 39: nit: rename either `ExpandPath` or `PrependPath`? having both as-is is a little misleading since one applies to the PATH env var and one doesn't

pkg/testutil/env/path.go
```diff
@@ -0,0 +1,51 @@
...
+	os.Setenv("PATH", fmt.Sprintf("%s:%s", filepath.Join(path...), os.Getenv("PATH")))
+}
+
+// ProjectDir returns the project root directory, determined by bazel. Panics if the directory can't be determined.
+func ProjectDir() string {
```
davidwding, line 44: does this need to be exported?

apis/v1alpha1/suite_test.go
```diff
@@ -54,24 +53,14 @@ func TestAPIs(t *testing.T) {
...
-
 	By("bootstrapping test environment")
-	testEnv = &envtest.Environment{
-		CRDDirectoryPaths: []string{filepath.Join("..", "..", "config", "crd", "bases")},
-	}
+	testEnv = env.NewEnv(runtime.NewSchemeBuilder(AddToScheme), filepath.Join("..", "..", "config", "crd", "bases"))
 
 	var err error
-	cfg, err = testEnv.Start()
+	testEnv.Start()
```
davidwding, line 62: why stop checking err here? delete line 64 if we want to keep it this way

apis/v1alpha1/suite_test.go
```diff
@@ -87,6 +76,5 @@ var _ = BeforeSuite(func(done Done) {
 
 var _ = AfterSuite(func() {
 	By("tearing down the test environment")
-	err := testEnv.Stop()
-	Expect(err).ToNot(HaveOccurred())
+	testEnv.Stop()
```
davidwding, line 79: same question as above re: discarding err

pkg/testutil/env/env.go
```diff
@@ -39,7 +38,17 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/envtest"
 )
 
+// testBinaries is used to set KUBEBUILDER_ASSETS variable to ensure things like etcd are available to all test
+// environments. This path is also added to the PATH.
```
davidwding, line 42: just curious - why do we need to both set `KUBEBUILDER_ASSETS` and also add to `PATH`?

davidwding (commented):
just nits and minor questions

apis/v1alpha1/suite_test.go
pseudomuto: Our implementation of `Env.Stop()` actually doesn't return an error. It prints out an error if one occurs, and panics if it fails to uninstall CRDs.

pkg/testutil/env/env.go
pseudomuto: When our test targets specify data resources like `//hack/bin:etcd`, it's typically for being able call `<thatpath>/etcd`. However, some of our packages require that the values be on the path, like _pkg/security_ for example. If this seems odd, we can separate the two and make two calls in `Env.Start`.

pkg/testutil/env/path.go
pseudomuto: Nope! Nice catch...will fix 👍

apis/v1alpha1/zz_generated.deepcopy.go
```diff
@@ -168,6 +168,13 @@ func (in *CrdbClusterSpec) DeepCopyInto(out *CrdbClusterSpec) {
 		*out = new(v1.Affinity)
 		(*in).DeepCopyInto(*out)
 	}
+	if in.AdditionalLabels != nil {
```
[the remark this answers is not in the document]
pseudomuto: Perhaps whenever `AdditionalLabels` was added we forgot to run `make dev/generate` ?

pkg/testutil/env/path.go
pseudomuto: Did a quick `s/PrependPath/PrependToPath/g`. Does that seem more clear?

pkg/testutil/env/env.go
davidwding: it does seem odd and it might be good to separate the two (if only to have a better understanding of what all the dependencies are), but we can do that later

davidwding (dismissed):

pkg/testutil/env/env.go
pseudomuto: Cool, I'll address it now just to avoid another TODO. Should be easy enough.

apis/v1alpha1/zz_generated.deepcopy.go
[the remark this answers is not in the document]
chrislovecnm: So this should be a test that the file is up to date.  I think verify-crds should cover this, but maybe I am wrong. Did you fix one of the generations through this PR?

e2e/openshift/openshift_packaging_test.go
```diff
@@ -109,9 +109,8 @@ func TestPackaging(t *testing.T) {
 	kubeconfig := os.Getenv("KUBECONFIG")
 	require.True(t, kubeconfig != "", "KUBECONFIG evn var not set")
 
-	// this call does some magic and ensures and sets
-	// oc in the PATH env variable for the test
-	paths.MaybeSetEnv("PATH", "oc", "hack", "bin", "oc")
+	// ensure oc will be on the path
```
chrislovecnm, line 112: can you add some comments around how bazel is pulling oc down and how it is set up?  Should we have it documented how this code is interacting with the binaries and how bazel pulls down those binaries?

pkg/testutil/env/path.go
```diff
@@ -0,0 +1,51 @@
...
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+)
+
+// ExpandPath expands a relative path to a workspace specific one.
```
chrislovecnm, line 27: There are two paths that this code follows and this code is critical to testing.  Can you elaborate on why we return one path vs another?  Are you noticed, this code does not work, testing does not work

chrislovecnm (commented):
thanks for the work. just a couple of comments

pkg/testutil/env/path.go
pseudomuto: We return the default (bazel test workspace relative) when running with bazel. There are times where it's nice to quickly run tests with go test, like within an editor for example. This allows that to work. I've updated the comments here to reflect that.

To be clear, in CI environments, and most cases really, the bazel path will be used. It seemed like a small concession to make this an option. Are you concerned that something will break here?

davidwding (approved):

pkg/testutil/env/path.go
chrislovecnm: Nope I like it, just need to document the hell out of the code like you did

chrislovecnm (approved):
LGTM

PR #686, opened 2021-08-23
add client-secure-operator.yaml to generated configs

rail:
Adds client-secure-operator.yaml to the list of generated files and its
content will be checked as a part of CI.


pseudomuto (approved):
This LGTM 👍

chrislovecnm (changes requested):
Can we add documentation here: https://github.com/cockroachdb/cockroach-operator/blob/master/DEVELOPER.md

About how this is working? I do not this is documented.

rail:
Good point. I'll update the PR.

Issue #545, opened 2021-05-28
[regression] installing a new cluster using the operator launches multiple vcheck jobs


keith-mcclellan:
I am seeing more than 1 vcheck job, even though the first job ran successfully and the crdb cluster started.  The second job does not terminate properly.  

steps to reproduce:
create a new cluster

![Screen Shot 2021-05-28 at 12 38 25 PM](https://user-images.githubusercontent.com/4730669/120015867-dfd00280-bfb1-11eb-8064-c08db3023b76.png)

[vcheck-multi-fire.log](https://github.com/cockroachdb/cockroach-operator/files/6562220/vcheck-multi-fire.log)

alinadonisa:
@keith-mcclellan  "crdb-tls-example-vcheck-27036997" is missing from the logs

udnay:
What's the job problem we're seeing in CI? Looking at the logs this might be another race condition. I see errors that the operator is unable to pull the DB's CR. In any case I don't think this is a release blocker, not sure how we test this though?

PR #690, opened 2021-08-26
Delete version checker job if it does not start

davidwding:
Fixes https://github.com/cockroachdb/cockroach-operator/issues/545

The issue is that if a vcheck job is pending for 2 minutes (for whatever reason, like lack of resources), a new vcheck job will be spun up every 2 minutes, leading to numerous pending vcheck jobs.

This PR fixes the immediate issue by deleting the vcheck job if it has not come up successfully.


pkg/resource/cluster.go
```diff
@@ -154,7 +154,7 @@ func (cluster Cluster) StatefulSetName() string {
 
 func (cluster Cluster) JobName() string {
 	slug.MaxLength = 63
-	return slug.Make(fmt.Sprintf("%s-vcheck-%d", cluster.Name(), getTimeHashInMinutes(time.Now())))
```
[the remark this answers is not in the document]
chrislovecnm: I think the best way to figure out if this works is e2e test or a unit test. We need to recreate that it is busted, and then test that we fixed it

chrislovecnm:
Does this fix our problem? Can you rebase please

davidwding:
> Does this fix our problem? Can you rebase please

Sorry about the lack of activity on this PR. Now that I've had some more time to generally read and understand the operator code, I'm confident that this fixes the reported bug and won't cause the potential race condition that you brought up.

davidwding:
Actually, on second thought, this might cause some undesired delays when trying to modify a cluster. Lemme think on it some more and I'll ping back when this is ready for another look.

davidwding:
@chrislovecnm @pseudomuto this is ready for another look. The version checker actor now deletes a job if the pod hasn't come up for 2 minutes. I added a test verifying that when version checker pods are unschedulable, we never end up with more than one.

pkg/actor/validate_version.go
```diff
@@ -139,6 +139,10 @@ func (v *versionChecker) Act(ctx context.Context, cluster *resource.Cluster) err
 		err := WaitUntilJobPodIsRunning(ctx, clientset, job, log)
 		if err != nil {
 			log.Error(err, "job not found")
+			if dErr := deleteJob(ctx, cluster, clientset, job); dErr != nil {
```
chrislovecnm, line 142: The log message does not make sense now

pkg/actor/validate_version.go
```diff
@@ -293,15 +300,9 @@ func (v *versionChecker) Act(ctx context.Context, cluster *resource.Cluster) err
...
-	//delete the job only if we have managed to get the version and we do not have any errors
-	err = clientset.BatchV1().Jobs(cluster.Namespace()).Delete(ctx, job.Name, metav1.DeleteOptions{
-		GracePeriodSeconds: ptr.Int64(5),
-		PropagationPolicy:  &dp,
-	})
-	if err != nil {
-		log.Error(err, "failed to delete the job")
+	// If we got here, the version checker job was successful. Delete it.
+	if dErr := deleteJob(ctx, cluster, clientset, job); dErr != nil {
+		log.Error(dErr, "failed to delete the job")
```
chrislovecnm, line 305: Does log.Error return an error? Wondering if we should return here

chrislovecnm (commented):
Just a nitpick on a log message. Otherwise looks great

pkg/actor/validate_version.go
davidwding: log.Error doesn't return an error, but do you think returning here is the right thing to do? I could see an argument either way - let the version checker finish and potentially have a dangling job, or just return and say the version checker failed.

pkg/actor/validate_version.go
davidwding: updating the error on the previous line to be more accurate

pkg/actor/validate_version.go
udnay-crl: If we error here I would worry that we can't move forward with the next step. Having a dangling job is probably ok IMO, better than having version checker come back as ok but fail on clean up. Maybe add more context in the error message. Check passed but delete failed.

davidwding: I've updated the error message as we just discussed. This PR should be good to go.

chrislovecnm (approved):
Lgtm

PR #693, opened 2021-08-30
consolidate handles logic

davidwding:
Consolidate all the Actor.Handles logic into a single func, Director.GetActionsToExecute, that Reconcile calls to determine which Actors should Act. Hopefully this gets us a bit closer to rewriting the operator as a state machine.

Added a number of unit tests to actor_test.go.

This PR does not intend to make any changes to business logic.


davidwding:
@pseudomuto before I get too far into this and write a bunch of tests and whatnot, could I get a check on whether you think this structure is a step in the right direction?

pseudomuto:
I like the idea here. One thing that stands out is that we're still running all of the actors on each event (assuming features and conditions are as expected). I wonder if there's a way we could avoid that. For example, if we're decommissioning, we don't need to bother with the version check and/or resizing PVs.

In any case, I think something like this is a step in the right direction as far as moving away from the actor model. We'd at least be able to have a single place to determine which actors to run.

davidwding:
Yeah, with all the actor Handles logic in one place it's a lot clearer how redundant all these actors are. We should definitely work on cleaning all that up after this.

pkg/actor/actor.go
```diff
@@ -72,15 +75,18 @@ func (e ValidationError) Error() string {
...
 
+type Director interface {
+	GetActionsToExecute(*resource.Cluster) []api.ActionType
+}
+
 // NewOperatorActions creates a slice of actors that control the actions or actors for the operator.
 // The order of the slice is critical so that the actors run in order, for instance update has to
 // happen before deploy.
-func NewOperatorActions(scheme *runtime.Scheme, cl client.Client, config *rest.Config) []Actor {
+func NewOperatorActions(scheme *runtime.Scheme, cl client.Client, config *rest.Config) map[api.ActionType]Actor {
```
pseudomuto, line 89: Rather than exposing this, what do you think about adding this map to `Director` and building it up in a constructor function? Then we can return actors to execute, rather than their types. It would also prevent the need for a reconciler to need actions as well as a director.

E.g.

```golang
type Director interface {
  GetActorToExecute(*resource.Cluster) []Actor
}

func NewDirector(scheme *runtime.Scheme, cl client.Client, config *rest.Config) Director {
  actions := // build map
  return &clusterDirector{scheme: scheme, client: cl, config: config, actions: actions}
}

type clusterDirector struct {
  ...
  actions map[api.ActionType]Actor
}
```

pkg/actor/actor_test.go
```diff
@@ -14,46 +14,168 @@ See the License for the specific language governing permissions and
...
-		require.True(t, xType.String() == expected, "expected:", expected, "got", xType.String())
-	}
+	utilfeature.DefaultMutableFeatureGate.Set("CrdbVersionValidator=false")
+	actions = director.GetActionsToExecute(cluster)
+	require.False(t, containsAction(actions, api.VersionCheckerAction))
+}
+
+func TestResizePVCFeatureGate(t *testing.T) {
+	// Setup fake client
+	cluster := testutil.NewBuilder("cockroachdb").
```
pseudomuto, line 80: Worth extracting a `createTestCluster` func?

pseudomuto: Another option might be to combine these tests in a table driven test like `TestDirectorGetActionsToExecute`. Then the cluster and director could be shared for all of them.

pkg/controller/cluster_controller.go
```diff
@@ -38,9 +38,10 @@ import (
 // ClusterReconciler reconciles a CrdbCluster object
 type ClusterReconciler struct {
 	client.Client
-	Log     logr.Logger
-	Scheme  *runtime.Scheme
-	Actions []actor.Actor
+	Log      logr.Logger
+	Scheme   *runtime.Scheme
+	Actors   map[api.ActionType]actor.Actor
```
pseudomuto, line 43: This is what I was talking about above. If Director knew about the actors map, the reconciler wouldn't have to.

pkg/actor/actor.go
davidwding: Sure, makes sense

davidwding (commented):
please take another look!

pkg/actor/actor.go
```diff
@@ -72,34 +75,89 @@ func (e ValidationError) Error() string {
...
-		newPartitionedUpdate(scheme, cl, config),
-		newResizePVC(scheme, cl, config),
-		newDeploy(scheme, cl, config, kube.NewKubernetesDistribution()),
-		newInitialize(scheme, cl, config),
-		newClusterRestart(scheme, cl, config),
+type Director interface {
+	GetActorsToExecute(*resource.Cluster) []Actor
+}
+
+type ClusterDirector struct {
```
pseudomuto, line 86: /nit not a show stopper, but I don't think this needs to be exported

pseudomuto (approved):

pseudomuto (approved):

PR #698, opened 2021-09-01
Add make commands for running locally

pseudomuto:
While testing out some validation locally, I noticed that our markers weren't correct for the VolumeClaim type. It expects, _**either**_ the spec or the source to be set, but not both. This means the both need to be marked optional.

In order to make this process easier to test, I've added `make dev/run` and `make dev/run-shell` targets. These run a local kind cluster (with the operator pre-installed) and a pre-configured shell respectively.

I've updated DEVELOPER.md with details about these targets.

**Terminal 1**

`$ make dev/run`

**Terminal 2**

```
$ make dev/run-shell
```


DEVELOPER.md
```diff
@@ -39,6 +39,28 @@ We have four different groups of commands withing the Makefile
...
+there's no need to set up a remote GKE/OpenShift cluster.
+
+[kind]: https://kind.sigs.k8s.io/
+
+**make dev/run**
+
+This command will get everything set up for you to begin testing out the operator. Specifically it will:
+
+* Start a kind cluster named test (context=kind-test)
+* Create the `dev` namespace
```
udnay, line 54: Why `dev` don't our examples assume `default` namespace?

udnay (dismissed):
LGTM just one question

DEVELOPER.md
pseudomuto: No particular reason, just a habit of mine. Think we should just use _default_ ?

udnay: My assumption would be that I could just copy paste the example and have it work and this breaks that assumption IMO. 

Where would you be using this?

PR #700, opened 2021-09-02
Fix inconsistency in required field

davidwding:
I noticed when reading through the code that the `Name` field has both `+required` and `omitempty`. Because of `omitempty` it doesn't actually end up with a required tag in the yaml.

Given that the entire `PodImage` struct is required in the CrdbCluster spec, I'm guessing that the original intent was for `Name` to be required. I removed `omitempty` and ran `make dev/generate`.


chrislovecnm:
Can you rebase so I can LGTM?

davidwding:
> Can you rebase so I can LGTM?

Done.

pseudomuto (approved):

PR #703, opened 2021-09-10
Dropping dev/run in favour of up/down

pseudomuto:
Currently, we've got the ability to run the operator locally using `make dev/run`. This starts a kind cluster and runs the operator in the foreground.

When we add webhooks for the admission controller (in progress), we'll need to ensure that the operator is deployed to the cluster rather than running in the foreground on the local host. This is necessary because the service will need to be able to use a selector (matchLabels) to find the pod that's running the webhook server.

This change adds `make dev/up` and `make dev/down` which start and stop the kind cluster respectively. When running `dev/up`, the image is published to gcr.io (`$DEV_REGISTRY`) and kind is configured with permissions to pull from there.

This is extracted from #699 in order to make the diff there smaller and more reviewable.


chrislovecnm:
I would like to start using a local registry, eventually

chrislovecnm (approved):
LGTM

PR #705, opened 2021-09-13
Use OpenAPI validation for node count

pseudomuto:
We currently check for the node count to be >=3 in the decommissioner. There's actually a `TODO` comment in there about adding a validation hook to avoid this. However, since the admission controller will always run the OpenAPI validation when resources are submitted to the API server, we don't actually need a hook for this.

I've added a minimum=3 marker to the spec, which will be enforced automatically, negating the need for this check.

**Proof:**

To try this out, you can update the `nodes` property in `examples/example.yaml` to be `2`. The run the following:

```
make dev/up && make dev/up-wait # start a local kind cluster
kubectl apply -f examples/example.yaml
=> The CrdbCluster "cockroachdb" is invalid: spec.nodes: Invalid value: 2: spec.nodes in body should be greater than or equal to 3

make dev/down
```

### DIFF
--- a/apis/v1alpha1/cluster_types.go
+++ b/apis/v1alpha1/cluster_types.go
@@ -32,6 +32,7 @@ import (
 type CrdbClusterSpec struct {
 	// Number of nodes (pods) in the cluster
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Number of nodes",xDescriptors="urn:alm:descriptor:com.tectonic.ui:podCount"
+	// +kubebuilder:validation:Minimum=3
 	// +required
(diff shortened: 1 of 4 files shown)


davidwding (commented):
Decommission doesn't allow a cluster to scale below 3, but we do we allow clusters with 1 node if they are initialized that way? This PR would break the second case if we do.

udnay:
We shouldn't allow that with the operator. The operator will break if we have a single node cluster.

davidwding:
Sounds reasonable. Is there anywhere we need to update documentation as well?

chrislovecnm (approved):
Lgtm

PR #706, opened 2021-09-13
Use local registry in dev

pseudomuto:
Rather than building and publishing images to GCR when running locally, we can use a local registry during development. Testing locally with `make dev/up`, this takes ~1m (on my machine) to get a functional cluster with the operator deployment available for use.


chrislovecnm (approved):
LGTM

PR #708, opened 2021-09-15
Adding kustomize macro so we can use it in bazel

pseudomuto:
We've currently got some things using kustomize, but others referencing yaml manifests directly. I think we can leverage kustomize to make things like changing the namespace we deploy to for example much simpler.

I've added the `kustomization` macro, which can be used to generate K8s manifests. These in turn can be used by `k8s_deploy`, `k8s_object(s)`, etc. as inputs.

To do this, I've updated kustomize to the latest version (4.3.0) in order to take advantage of
[replacements](https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/replacements/). These are used to set the namespaceSelector for the webhooks for example, by referencing the namespace from the service. This allows us to set the namespace once in kustomization.yaml and have it used for all resources (unless otherwise specified).

Finally, I've updated the testenv to use these new resources. This ensures that both the crds, and the webhooks are running in the testenv.

**Example**

```bazel
load("@k8s_deploy//:defaults.bzl", "k8s_deploy")
load("//hack/build:kustomize.bzl", "kustomization")

// generates manifest.yaml at runtime by processing kustomization.yaml
kustomization(
    name = "manifest",
    srcs = [":all-srcs"],
)

// deploys the result from :manifest
k8s_deploy(
    name = "my-resource",
    template = ":manifest",
    visibility = ["//visibility:public"],
)
...
...
```
If you'd like to verify the results, you can run `bazel run //<path>:manifest.preview. The resulting manifest will be printed to stdout.

### DIFF
--- a/apis/v1alpha1/BUILD.bazel
+++ b/apis/v1alpha1/BUILD.bazel
@@ -37,12 +37,6 @@ go_test(
         "volume_test.go",
         "webhook_test.go",
     ],
-    data = [
-        "//config/crd/bases:crd_manifest",
-        "//hack/bin:etcd",
-        "//hack/bin:kube-apiserver",
-        "//hack/bin:kubectl",
-    ],
     deps = [
         ":go_default_library",
         "//pkg/client/clientset/versioned:go_default_library",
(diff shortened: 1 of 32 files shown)


chrislovecnm (approved):
LGTM

PR #710, opened 2021-09-16
Added Custom Annotations support in crdb cluster #696

### LINKED ISSUE #696: Implement Custom Annotations in CR
Provide functionality in the CR to the end user to be able to annotate clusters  for 3rd party service integration

abhishekdwivedi3060:
Closes #696 

* Added Custom Annotations support in crdb cluster
* Added the new API field "additionalAnnotations" for adding user defined annotations to CRDB cluster resources
* Added/Modified existing unit test for this change

### DIFF
--- a/apis/v1alpha1/cluster_types.go
+++ b/apis/v1alpha1/cluster_types.go
@@ -109,6 +109,10 @@ type CrdbClusterSpec struct {
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Map of additional custom labels"
 	// +optional
 	AdditionalLabels map[string]string `json:"additionalLabels,omitempty"`
+	// (Optional) Additional custom resource annotations that are added to all resources
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Map of additional custom annotations"
+	// +optional
+	AdditionalAnnotations map[string]string `json:"additionalAnnotations,omitempty"`
 	// (Optional) Tolerations for scheduling pods onto some dedicated nodes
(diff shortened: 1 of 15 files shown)


pkg/resource/statefulset.go
```diff
@@ -191,6 +194,8 @@ func (b StatefulSetBuilder) makePodTemplate() corev1.PodTemplateSpec {
 	pod := corev1.PodTemplateSpec{
 		ObjectMeta: metav1.ObjectMeta{
 			Labels: b.Selector,
```
chrislovecnm, line 196: Does this cause pod restart? Does datadog, for instance, need annotations on the pods?

abhishekdwivedi3060: Yes, it does. Change it sts spec triggers restart.
Also, most of the monitoring tools like Datadog, Prometheus works on pod annotations only, so added custom annotations in the pod meta as well

apis/v1alpha1/cluster_types.go
```diff
@@ -109,6 +109,10 @@ type CrdbClusterSpec struct {
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Map of additional custom labels"
 	// +optional
 	AdditionalLabels map[string]string `json:"additionalLabels,omitempty"`
+	// (Optional) Additional custom resource annotations that are added to all resources
```
pseudomuto, line 112: Think we should add something here to remind users that changing this will restart the cluster?

pseudomuto (dismissed):

apis/v1alpha1/cluster_types.go
abhishekdwivedi3060: Done

pseudomuto (approved):

PR #714, opened 2021-09-16
Bump build dependencies

pseudomuto:
Some of our build dependencies were getting quite old, and some were no longer needed. I went through each one and verified first if it was needed, then updated it when necessary.

**EDIT:** I initially intended to bump the build deps, then follow up with a PR for normal runtime deps. However, some of the build dependency updates require updated runtime dependencies. For example, bumping etcd to 3.5.0 requires an updated version of controller-runtime (which in turn requires an updated version of client-go). So I've updated the strictly necessary runtime deps here as well.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -32,10 +32,10 @@ http_archive(
 
 http_archive(
     name = "io_bazel_rules_go",
-    sha256 = "69de5c704a05ff37862f7e0f5534d4f479418afc21806c887db544a316f3cb6b",
+    sha256 = "8e968b5fcea1d2d64071872b12737bbb5514524ee5f0a4f54f5920266c261acb",
     urls = [
-        "https://mirror.bazel.build/github.com/bazelbuild/rules_go/releases/download/v0.27.0/rules_go-v0.27.0.tar.gz",
-        "https://github.com/bazelbuild/rules_go/releases/download/v0.27.0/rules_go-v0.27.0.tar.gz",
+        "https://mirror.bazel.build/github.com/bazelbuild/rules_go/releases/download/v0.28.0/rules_go-v0.28.0.zip",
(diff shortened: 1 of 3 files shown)


davidwding:
It looks like 4 of the e2e tests are failing? Glancing at the logs it's not clear to me whether it's caused by this PR or if all of them just happened to time out.

pseudomuto:
I'll have to dig into this a bit. At first glance it seems that the way we find the K8s distro by checking the kubelet version no longer works for kind. I'll verify and see what our options are.

pseudomuto:
So, I've confirmed it's not an issue with the PDB, but rather with the readiness probes (which cause the PDB to continue reconciling).

The probe now fails with a 503 for /health?ready=1 for pods. Still digging...

pseudomuto:
I've backed out the non-build-related dependencies to get this passing. I'll follow up with another PR for the controller-runtime related changes

pkg/actor/deploy.go
```diff
@@ -123,27 +123,23 @@ func (d deploy) Act(ctx context.Context, cluster *resource.Cluster) error {
 		return nil
 	}
 
-	// if we only have one Node we cannot have a PDB
-	// TODO we need to validate this in the CRD API
-	if cluster.Spec().Nodes > 1 {
```
pseudomuto, line 128: Admission controller guarantees that this count >=2

davidwding (approved):

pkg/actor/deploy.go
chrislovecnm: Should we have this in a different PR?

pseudomuto:
Backed out the change to deploy action (see #721)

davidwding (approved):

PR #716, opened 2021-09-20
Added go-lint ci job and make command #711

### LINKED ISSUE #711: Add a go lint CI job
* Add a go lint CI job for making sure the code is well-formatted as per different checks like goimports, gofmt etc

abhishekdwivedi3060:
Closes #711
* Added go-lint ci job
* Added make command for dev


hack/BUILD.bazel
```diff
@@ -183,6 +185,17 @@ sh_test(
     ],
 )
 
+sh_binary(
+    name = "update-golangci-lint",
```
pseudomuto, line 189: /nit since we're not updating anything, I think we should remove the `update-` prefix here (for the script as well).

hack/bin/deps.bzl
```diff
@@ -223,6 +224,39 @@ def install_kubectl():
         urls = ["https://storage.googleapis.com/kubernetes-release/release/v1.18.0/bin/linux/amd64/kubectl"],
     )
 
+# Define rules for different golangci-lint versions
+def install_golangci_lint():
+    http_archive(
+        name = "golangci_lint_1_42_darwin",
```
pseudomuto, line 230: Including the version here means that when we update this, we need to do it in two places (here and in BUILD.bazel). Can we omit the version (i.e. `golangci_lint_darwin`)?

hack/update-golangci-lint.sh
```diff
@@ -0,0 +1,42 @@
...
+    bazel run //hack:update-golangci-lint
+  )
+  exit 0
+fi
+
+golangci_lint=$(realpath "$1")
+
+cd "$BUILD_WORKSPACE_DIRECTORY"
+
+export GO111MODULE=on
```
pseudomuto, line 39: We're using go 1.16 so I don't think this is necessary

hack/BUILD.bazel
abhishekdwivedi3060: Removed prefix

hack/bin/deps.bzl
abhishekdwivedi3060: Yes, makes sense. Removed

hack/update-golangci-lint.sh
abhishekdwivedi3060: Removed export statement from this file. 
However, that statement is currently used in multiple files. Might need to remove from those files as well.

pseudomuto (approved):

pseudomuto (approved):

hack/BUILD.bazel
```diff
@@ -183,6 +185,17 @@ sh_test(
     ],
 )
 
+sh_binary(
```
chrislovecnm, line 188: We need a call to test and also a call to fix the problems. I am only seeing one new target. So when we run bazel test //hack/... We run a test.

chrislovecnm (changes requested):
See comment

hack/BUILD.bazel
abhishekdwivedi3060: Correct me if I understood it wrong, you are saying to add a target to fix the lint problems something like `gofmt -w`.
But golangci-lint issues can't be solved directly, those needs to be fixed manually by the developer.
Also, if we add a test right now, it's going to fail as there are many golangci-lint in master at the moment

abhishekdwivedi3060: @chrislovecnm Thoughts?

Makefile
```diff
@@ -210,6 +210,10 @@ dev/build:
 dev/fmt:
 	bazel run //hack:update-gofmt
 
+.PHONY: dev/fmt
+dev/golangci-lint:
+	bazel run //hack:golangci-lint
```
pseudomuto, line 215: If we changed this to `bazel run //hack/bin:golangci-lint` do we need the bash script?

hack/BUILD.bazel
pseudomuto: I _think_ what he's referring to is to have the ability to run `bazel test //hack/...` and have the linter run.

```bazel
sh_test(
  name = "verify-golangci-lint",
  srcs = ["//hack/bin:golangci-lint"],
  data = ["@//:all-srcs"],
)
```

If we do that, we don't need a separate GH action, since `make test/all` (which CI runs) will include the tests from `//hack`

pkg/kuberecord/recorder_test.go
```diff
@@ -34,7 +35,11 @@ func TestRecorder(t *testing.T) {
 
 	cfg, err := env.Start()
 	require.NoError(t, err)
-	defer env.Stop()
+	defer func(env *envtest.Environment) {
+		if err := env.Stop(); err != nil {
+			log.Error()
```
pseudomuto, line 40: Can we `t.Logf` here to avoid prometheus?

hack/verify-golangci-lint.sh
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env bash
```
pseudomuto, line 1: Is this script still used?

pkg/clustersql/settings_test.go
```diff
@@ -143,6 +145,7 @@ func TestRangeMoveDuration(t *testing.T) {
 					{Target: "zone-1", Config: ZoneConfig{RangeMaxBytes: 20}},
 					{Target: "zone-2", Config: ZoneConfig{RangeMaxBytes: 30}},
 				},
+				// TODO: Is it required. Output will always be 0s because of interger division
```
pseudomuto, line 148: As discussed in Slack, we should open an issue here. This is passing because we have the same issue in the code under test. It might be the case that these perfect numbers don't exist in reality, but to be sure we should take a closer look.

For now, I think leaving these comments in place and adding an issue to track should be sufficient.

hack/verify-golangci-lint.sh
abhishekdwivedi3060: No, it is no longer required. Removed

pkg/kuberecord/recorder_test.go
abhishekdwivedi3060: Done

hack/crdbversions/main.go
```diff
@@ -62,6 +62,8 @@ type templateData struct {
 // readCrdbVersions reads CRDB versions from a YAML file and sorts them
 // according to the semantic version sorting rules
 func readCrdbVersions(r io.Reader) ([]*semver.Version, error) {
+	// nolint
```
pseudomuto, line 65: Is this nolint marker necessary? We're using 1.6

pkg/testutil/require.go
```diff
@@ -67,7 +66,7 @@ func RequireClusterToBeReadyEventuallyTimeout(t *testing.T, sb testenv.DiffingSa
 
 		if !statefulSetIsReady(ss) {
 			t.Logf("stateful set is not ready")
-			logPods(context.TODO(), ss, cluster, sb, t)
+			_ = logPods(context.TODO(), ss, cluster, sb, t)
```
pseudomuto, line 69: if logPods can return an error, maybe we should t.Logf it?

pkg/testutil/validate_headers_test.go
```diff
@@ -116,14 +116,14 @@ func TestValidate(t *testing.T) {
 		t.Fatal("error running Validate", err)
 	}
 
-	for _, filename := range *non {
-		t.Logf("nonvalid file: %s", filename)
-	}
-
 	if non == nil {
 		t.Fatal("test did not find any bad files")
```
pseudomuto, line 120: This doesn't seem right (no your changes), but why would we fail if there were no bad files?

pkg/testutil/validate_headers_test.go
```diff
@@ -116,14 +116,14 @@ func TestValidate(t *testing.T) {
...
 	}
 
+	for _, filename := range *non {
+		t.Logf("nonvalid file: %s", filename)
```
pseudomuto, line 124: nit: s/nonvalid/invalid/

pseudomuto (approved):
Nothing but nits here. Overall this LGTM 👍

pkg/testutil/validate_headers_test.go
abhishekdwivedi3060: In the test-case, we are creating some good files and some bad files initially as a test data and it is expected from `Validate()` func that it should return `nonconformingFilesPtr`. So if `non==nil` it means our code is not catching bad files and   hence we are failing

pkg/testutil/require.go
abhishekdwivedi3060: Done

hack/smoketest/main.go
```diff
@@ -23,6 +23,7 @@ import (
 	"path/filepath"
 	"time"
 
+	log "github.com/sirupsen/logrus"
```
pseudomuto, line 26: nit: Do we need the logging here on top of the `fmt.Print` statements? If we remove this (or at least use logr) we'd drop the the new dependency added to go.mod

pseudomuto (approved):

PR #718, opened 2021-09-20
don't requeue permanent errors

davidwding:
I'm doing an audit of the way the actors handle errors, and as part of that, looked into how reconciliation handles the errors. It doesn't really make sense that a "permanent" error should get ever get requeued, so let's make them not requeue.

The two existing uses of PermanentErr are indeed pretty permanent: both occur when the `cockroach version` command doesn't run successfully.


pseudomuto (approved):
If these are not transient, this seems legit to me 👍

PR #720, opened 2021-09-21
Removing some unused or unnecessary hack scripts

pseudomuto:
We have a bunch of targets in //hack that check if they're running in bazel and either warn the user to install bazel or subsequently call itself via bazel run/test ....

If we remove the scripts (where possible), we can avoid the need to have these checks, as well as configuring an explicit target in the BUILD files. For most of these, I was able to simply add a new make target and use `bazel run //hack/bin:<cmd> ...`.


hack/BUILD.bazel
```diff
@@ -129,60 +79,20 @@ sh_test(
...
-    srcs = ["update-gofmt.sh"],
-    args = [
-        "$(location %s)" % GOFMT,
-    ],
-    data = [
-        GOFMT,
-    ],
-)
-
-sh_test(
```
chrislovecnm, line 174: Why are we removing this?

pkg/actor/actor.go
```diff
@@ -18,6 +18,7 @@ package actor
 
 import (
 	"context"
```
chrislovecnm, line 20: Can you remove this please?

pseudomuto: You mean the new line? The new `make dev/fmt` command added that.

hack/BUILD.bazel
pseudomuto: It's covered with `make dev/lint`

hack/BUILD.bazel
```diff
@@ -129,58 +79,29 @@ sh_test(
...
-    args = [
-        "$(location %s)" % GOFMT,
-    ],
-    data = [
-        GOFMT,
-    ],
-)
-
 sh_test(
     name = "verify-gofmt",
```
pseudomuto, line 97: It's back so that `bazel test //hack/...` runs gofmt

chrislovecnm (approved):
LGTM

PR #722, opened 2021-09-21
make actors atomic

davidwding:


chrislovecnm:
I know this is in draft, but how will this work with multiple CR databases and the operator?

davidwding:
> I know this is in draft, but how will this work with multiple CR databases and the operator?

When I get this fully working, it should just work with multiple CR databases, since each individual CrdbCluster has its own status, which is what the operator uses to lock per cluster.

config/crd/bases/crdb.cockroachlabs.com_crdbclusters.yaml
```diff
@@ -1089,6 +1089,9 @@ spec:
           status:
             description: CrdbClusterStatus defines the observed state of Cluster
             properties:
+              activeActor:
+                description: ActiveActor keeps track of which Actor is currently running
```
chrislovecnm, line 1093: I think we want this in an annotation or status rather than a field for the crd.  Unless a user needs to modify this field.

chrislovecnm (changes requested):

pkg/resource/cluster.go
```diff
@@ -110,6 +110,20 @@ func (cluster Cluster) SetActionUnknown(atype api.ActionType) {
...
+	now := metav1.NewTime(time.Now().Round(time.Second))
+
+	cluster.cr.Status.DirectorState = state
+	cluster.cr.Status.DirectorStateUpdatedAt = now
+	cluster.cr.Status.DirectorObservedGeneration++
+
+	return now, cluster.cr.Status.DirectorObservedGeneration
+}
+func (cluster Cluster) SetActiveActor(actor string) {
+	cluster.cr.Status.ActiveActor = actor
```
pseudomuto, line 124: Should setting the actor also update the director state?

pkg/actor/actor.go
```diff
@@ -161,6 +184,69 @@ func (cd *clusterDirector) GetActorsToExecute(cluster *resource.Cluster) []Actor
...
+	currentState := status.DirectorState
+	switch currentState {
+	case DirectorStateAvailable:
+		break
+	case DirectorStateBusy:
+		currentStateUpdatedAt := cluster.Status().DirectorStateUpdatedAt
+		if currentStateUpdatedAt.Add(maxTimeBusy).Before(time.Now()) {
+			return PermanentErr{Err: errors.New("director has timed out")}
+		} else {
+			return DirectorLockError{Err: errors.Newf("an actor is currently performing an action: %s", status.ActiveActor)}
```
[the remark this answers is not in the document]
davidwding: Taking this discussion into Slack

pkg/actor/actor.go
```diff
@@ -161,6 +184,69 @@ func (cd *clusterDirector) GetActorsToExecute(cluster *resource.Cluster) []Actor
...
+		if err := fetcher.Fetch(cr); err != nil {
+			return PermanentErr{Err: fmt.Errorf("could not retrieve cluster: %w", err)}
+		}
+		refreshedCluster := resource.NewCluster(cr)
+		status = refreshedCluster.Status()
+
+		// It's possible that the newly retrieved cluster does not yet reflect the update just made. If this is the case,
+		// try again.
+		if status.DirectorObservedGeneration < newGeneration {
+			continue
```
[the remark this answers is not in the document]
davidwding: I'll add a 5 second sleep for now so it's not so tight

pkg/actor/actor.go
```diff
@@ -161,6 +184,69 @@ func (cd *clusterDirector) GetActorsToExecute(cluster *resource.Cluster) []Actor
...
+		}
+
+		if status.DirectorState != DirectorStateBusy || !status.DirectorStateUpdatedAt.Equal(&directorUpdatedAt) {
+			return PermanentErr{Err: fmt.Errorf("active director lost lock; this should not have happened")}
```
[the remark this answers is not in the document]
davidwding: It's not actually clear to me what should be done here! It seems pretty fatal to acquire an exclusive lock then mysteriously lose it. Maybe we can suggest restarting the operator?? WDYT?

pkg/resource/cluster.go
davidwding: I'll combine UpdateDirectorState and SetActiveActor since we're always calling them together

config/crd/bases/crdb.cockroachlabs.com_crdbclusters.yaml
chrislovecnm: Doh! Correct

davidwding:
Now that one actor runs per reconcile, the benefit of this is much less clear and it just adds extra overhead to reason about. Closing.

PR #724, opened 2021-09-22
Add labels to PVCs

pseudomuto:
Our intention with `AdditionalLabels` is to add them to all resources created by CrdbCluster. Currently, the PVCs are missing these. So I've updated the StatefulSet resource appropriately to add the labels.

How to verify this works:

```shell
$ make dev/up
$ kubectl get pods -w # wait until at least one of the db pods is running and ready
$ k get pvc datadir-cockroachdb-0 -oyaml
# verify that `crdb: is-cool` is in the labels
$ make dev/down
```

### DIFF
--- a/pkg/resource/statefulset.go
+++ b/pkg/resource/statefulset.go
@@ -39,14 +39,15 @@ const (
 	grpcPortName = "grpc"
 	sqlPortName  = "sql"
 
-	DataDirName      = "datadir"
-	DataDirMountPath = "/cockroach/cockroach-data/"
+	dataDirName      = "datadir"
+	dataDirMountPath = "/cockroach/cockroach-data/"
 
 	certsDirName = "certs"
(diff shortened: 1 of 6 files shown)


pkg/resource/statefulset.go
```diff
@@ -85,10 +86,11 @@ func (b StatefulSetBuilder) Build(obj client.Object) error {
 		Template: b.makePodTemplate(),
 	}
 
-	if err := b.Spec().DataStore.Apply(DataDirName, DbContainerName, DataDirMountPath, &ss.Spec,
+	if err := b.Spec().DataStore.Apply(dataDirName, DbContainerName, dataDirMountPath, &ss.Spec,
 		func(name string) metav1.ObjectMeta {
 			return metav1.ObjectMeta{
-				Name: DataDirName,
+				Name:   dataDirName,
+				Labels: b.Spec().AdditionalLabels,
```
chrislovecnm, line 93: Should we do the full label set?

pseudomuto:
Updated to include all labels rather than just AdditionalLabels

chrislovecnm (approved):
LGTM

PR #726, opened 2021-09-23
Add correct namespace to generated webhook cert

### LINKED ISSUE #725: Webhook namespace is hard-coded
The new webhook namespace cert is hard-coded to `default`. There should be a way to pass in the namespace to the cert generation function.

https://github.com/cockroachdb/cockroach-operator/blob/4e7d708087b08016eb014f03135e0b19d7b2a49c/pkg/resource/webhook_secret.go#L177-L185

mbrancato:
This passes the namespace of the operator to the DNS names used in the webhook TLS certificate.

Fixes #725

### DIFF
--- a/pkg/resource/webhook_secret.go
+++ b/pkg/resource/webhook_secret.go
@@ -24,6 +24,7 @@ import (
 	"crypto/x509"
 	"crypto/x509/pkix"
 	"encoding/pem"
+	"fmt"
 	"math/big"
 	"time"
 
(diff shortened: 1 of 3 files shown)


pkg/resource/webhook_secret.go
```diff
@@ -86,7 +87,8 @@ func CreateWebhookSecret(ctx context.Context, client v1.SecretInterface) (*Webho
 	}
 
 	log.V(debugLevel).Info("Generating server certificate")
-	tlsCert, tlsKey, err := generateCert(ca, caPriv)
+	print(client)
```
pseudomuto, line 90: Forgot to remove this?

pseudomuto (changes requested):
Thanks for the PR @mbrancato! This LGTM, just need to remove the print statement then we can get this merged.

pkg/resource/webhook_secret.go
mbrancato: yep, done now.

pseudomuto (approved):

PR #728, opened 2021-09-23
Tear down test env with each test

davidwding:
There's an issue with the existing test structure where the controller may log something in between when a test finishes and when the test env is torn down, causing a panic. I'm working on a separate PR that adds some more logging to the controller and this race condition became really apparent.

Fix by starting and stopping the test env tightly around each test.

### DIFF
--- a/e2e/versionchecker/versionchecker_test.go
+++ b/e2e/versionchecker/versionchecker_test.go
@@ -18,7 +18,6 @@ package versionchecker_test
 
 import (
 	"flag"
-	"os"
 	"testing"
 	"time"
 
@@ -39,18 +38,6 @@ import (
 // We may have a threadsafe problem where one test starts messing with another test
 var parallel = *flag.Bool("parallel", false, "run tests in parallel")
 
-// TODO should we make this an atomic that is created by env pkg?
-var env *testenv.ActiveEnv
-
-// TestMain wraps the unit tests. Set TEST_DO_NOT_USE_KIND environment variable to any value
-// if you do not want this test to start a k8s cluster using kind.
-func TestMain(m *testing.M) {
-	e := testenv.CreateActiveEnvForTest()
-	env = e.Start()
-	code := testenv.RunCode(m, e)
-	os.Exit(code)
-}
-
 // TestVersionCheckerJobPodPending creates a cluster and then tries to run an unscheduleable version checker job.
 // It checks that there is never more than one version checker job running.
 func TestVersionCheckerJobPodPending(t *testing.T) {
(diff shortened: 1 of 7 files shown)


e2e/versionchecker/versionchecker_test.go
```diff
@@ -85,12 +77,6 @@ func TestVersionCheckerJobPodPending(t *testing.T) {
 				testutil.RequireAtMostOneVersionCheckerJob(t, sb, 300*time.Second)
 			},
 		},
-		{
-			Name: "teardown",
```
pseudomuto, line 89: Was this removed on purpose?

pkg/testutil/env/env.go
```diff
@@ -139,13 +137,6 @@ func CreateActiveEnvForTest() *Env {
 	return NewEnv(runtime.NewSchemeBuilder(api.AddToScheme))
 }
 
-func RunCode(m *testing.M, e *Env) int {
```
pseudomuto, line 142: Since all of the e2e tests use the same setup, perhaps we could add something like this (no biggie though):

```go
func WithActiveEnv(fn func (*Env)) {
  env := CreateActiveEnv()
  defer env.Stop()

  fn(env.Start())
}
```

pseudomuto (commented):
This LGTM, just a question about a deleted step and whether or not we should make a new helper

e2e/versionchecker/versionchecker_test.go
davidwding: Yeah, this was an earlier attempt to fix the issue when I had a less good understanding of what exactly the issue was.

pkg/testutil/env/env.go
davidwding: `defer` needs to be called in each Test* func itself - with this it'll get called right after WithActiveEnv returns. I think I can make this a little bit cleaner though - please hold

davidwding: Oh wait, I think I see what you're getting at, lemme see if I can make that work

pseudomuto: To be clear, I was thinking the tests would use this. Like:

```go
func TestMyThing(t *testing.T) {
  ...
  ...
  env.WithActiveEnv(func(env *env.Env) {
    ...tests here
  })
}
```

Honestly, it's not a big deal. If that feels over-engineered we can skip it for now

davidwding: Yeah, I think that feels a little much, but if env setup gets more complex I think doing this would be good. Let's just keep it as-is for now?

pseudomuto (approved):

PR #729, opened 2021-09-23
fix more flakiness in e2e tests

davidwding:
Some of our e2e tests involve getting the CrdbCluster, changing something in its spec, and updating the CrdbCluster. If anything changes in the cluster between the call to Get and Update, there is a write conflict and the test fails, which I've observed a few times.

Fix all these flaky tests by using Patch instead of Update. Patch will always patch in a calculated diff, even if CrdbCluster changes after being fetched.


chrislovecnm (changes requested):
LGTM

chrislovecnm (approved):

PR #733, opened 2021-09-27
actually update cluster conditions in initialize actor

davidwding:
It looks like the intent of this one existing line in Initialize is to actually set the cluster initialized condition to true, but currently it's only working by coincidence when the updated cluster gets written way downstream for something unrelated. The bit of refactoring I've done so far has surfaced this bug.

Make it so we actually update the cluster initialized condition within the initialize actor.

### DIFF
--- a/pkg/actor/initialize.go
+++ b/pkg/actor/initialize.go
@@ -137,7 +137,7 @@ func (init initialize) Act(ctx context.Context, cluster *resource.Cluster) error
 			return errors.Wrap(err, msg)
 		}
 		refreshedCluster := resource.NewCluster(newcr)
-		refreshedCluster.SetTrue(api.InitializedCondition)
+		refreshedCluster.SetTrue(api.CrdbInitializedCondition)
 
 		err = init.client.Status().Update(ctx, refreshedCluster.Unwrap())
 		if err != nil {
--- a/pkg/actor/initialize.go
+++ b/pkg/actor/initialize.go
@@ -128,17 +128,23 @@ func (init initialize) Act(ctx context.Context, cluster *resource.Cluster) error
 		return errors.Wrap(err, msg)
 	}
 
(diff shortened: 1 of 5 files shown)


pkg/actor/initialize.go
```diff
@@ -127,9 +128,32 @@ func (init initialize) Act(ctx context.Context, cluster *resource.Cluster) error
 		return errors.Wrap(err, msg)
 	}
 
-	cluster.SetTrue(api.InitializedCondition)
+	fetcher := resource.NewKubeFetcher(ctx, cluster.Namespace(), init.client)
+	err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
chrislovecnm, line 132: How are we getting conflicts?  I don't disagree that we should retry, but how would we need to?

pkg/actor/initialize.go
```diff
@@ -127,9 +128,32 @@ func (init initialize) Act(ctx context.Context, cluster *resource.Cluster) error
 		return errors.Wrap(err, msg)
 	}
 
-	cluster.SetTrue(api.InitializedCondition)
+	fetcher := resource.NewKubeFetcher(ctx, cluster.Namespace(), init.client)
```
chrislovecnm, line 131: This is pretty important code, should we add some code-level comments?

chrislovecnm (commented):
More comments and nit picks ... otherwise looks great!

pkg/actor/initialize.go
davidwding: Added a comment giving an example of how this might happen

pkg/actor/initialize.go
davidwding: Done, let me know if you'd like anything else in particular clarified with a comment

pseudomuto (approved):
👍 for the addition of the comments

PR #737, opened 2021-09-28
Update CRDB versions

pseudomuto:
Closes #694 

I've updated the CRDB versions here and excluded v21.1.8 which prevented downgrades.


davidwding (approved):

examples/example.yaml
```diff
@@ -40,9 +40,9 @@ spec:
       memory: 8Gi
   tlsEnabled: true
 # You can set either a version of the db or a specific image name
-# cockroachDBVersion: v21.1.7
+# cockroachDBVersion: v21.1.9
   image:
-    name: cockroachdb/cockroach:v21.1.7
+    name: cockroachdb/cockroach:v21.1.9
```
chrislovecnm, line 45: Did you modify the template that generates this file? Most of our YAML files are generated to update with the latest version.

hack/update-crdb-versions.sh
```diff
@@ -47,8 +51,14 @@ cat > crdb-versions.yaml << EOF
 
 CrdbVersions:
 EOF
+}
+
+dump_versions() {
+  local nameFilter=".data[] .repositories[] .tags[] .name"
+  # Skip unsupported versions and the latest tag
+  for version in $(curl -s $URL | jq -r "${nameFilter}" | grep -v ^v19 | grep -v ^v21.1.8$ | grep -v latest | grep -v ubi$ | sort --version-sort); do
```
chrislovecnm, line 59: This is a pretty complicated command that can fail a lot of ways.  Including jq is not installed.  Should we break this into more steps and check that the data is returned is valid?

chrislovecnm (changes requested):
See comments

hack/update-crdb-versions.sh
pseudomuto: I agree about breaking this up. However, this is in place now. Meaning jq is installed in the CI environment where the PR bot runs.

I was talking with @davidwding earlier about converting some of these critical bash scripts into some nice, testable, go apps. A problem for another PR though.

udnay: I agree, I think we can file an issue and ship this

pseudomuto: Issue logged https://github.com/cockroachdb/cockroach-operator/issues/738

davidwding (approved):

hack/update-crdb-versions.sh
pseudomuto: @chrislovecnm you good with this now?

chrislovecnm (approved):

chrislovecnm:
LGTM

davidwding (approved):

PR #739, opened 2021-09-28
Make operator logs more debuggable

davidwding:
This PR is a lot of files but fundamentally has just two components:
- add a unique `ReconcileId` to the logger constructed with each new call to `Reconcile`
- propagate `ReconcileId` throughout the entire call to `Reconcile`, including down to the Actors

The purpose of this is to make the operator easier to debug. Currently, it is very difficult to tell where one call to `Reconcile` ends and where another begins in the logs (and eventually they may even interleave). The hope is that by attaching a `ReconcileId` trace to all logs in one `Reconcile` call, issues will be easier to debug, both in the e2e tests and with a live-running operator.


.bazelrc
```diff
@@ -1,2 +1,2 @@
 build --workspace_status_command hack/build/print-workspace-status.sh
-test --test_output=errors --test_output=streamed
+test --test_output=streamed
```
chrislovecnm, line 2: This will make the unit tests run single threaded. Can we have a dev bazel profile?

chrislovecnm (commented):
Just a single nit pick

chrislovecnm:
Missed that this is a draft. Opps. Love the changes!

.bazelrc
davidwding: I'll just make it `errors`, I don't think there's any compelling reason for it to be `streamed`. Even running locally I find the output from `errors` to be sufficient.

pseudomuto (approved):

PR #742, opened 2021-09-29
fix CertificateGenerated annotation and associated TODO

davidwding:
There has been a long-standing TODO in actor.go saying that the CertificateGenerated condition on CrdbClusterStatus was not working. The issue was that the condition was not being written to the object correctly. Fix the bug and clean the code up slightly.

To be extremely specific, when updating CrdbClusterStatus from the operator, one needs to call `rc.client.Status().Update(ctx, crdbobj)` and **not** `rc.client.Update(ctx, crdbobj)`. The latter will not work. I was bitten by this on a separate issue yesterday.

Updated the actor unit tests as well.


davidwding:
I believe the failure in e2e-kind-create will be addressed by https://github.com/cockroachdb/cockroach-operator/pull/744

pkg/actor/actor_test.go
```diff
@@ -106,54 +107,65 @@ func TestClusterRestartFeatureGate(t *testing.T) {
...
-
-	utilfeature.DefaultMutableFeatureGate.Set("UseDecommission=true,CrdbVersionValidator=true,ResizePVC=true,ClusterRestart=true")
-
-	actors := director.GetActorsToExecute(cluster)
-	require.True(t, actorsHaveTypes(actors, []api.ActionType{api.VersionCheckerAction, api.RequestCertAction}))
+	return types
 }
 
-func TestVersionCheckedButNotInitialized(t *testing.T) {
+func TestAllConditionCombinations(t *testing.T) {
```
pseudomuto, line 118: 👍

pseudomuto (approved):

Issue #738, opened 2021-09-28
Convert Some Scripts to Go


pseudomuto:
We have a number of scripts that we use to update files used to release the operator. These are critical steps that are sometimes complex. Rather than using bash, I think we should convert some of these to testable go apps.

Perhaps we could start with something simple, like _hack/update-crdb-versions.sh_.

SumLare:
How urgent is this? I can try to tackle this today/tomorrow

pseudomuto:
Hey @SumLare, thanks so much! Not something that we need to do ASAP, but a direction we want to move towards. Any help would be much appreciated.

PR #746, opened 2021-09-30
Adding make release/new

pseudomuto:
Currently we have one process for patch releases, and a separate one for others. There doesn't seem to be any reason for this, at least not anymore.

With this PR, the new process is simplified to running a single make command, regardless of which type of release you're doing.

Sort of related to #738 

# How it works

Running `make release/new VERSION=<desired-version>` will:

* Validate the version matches `N.N.N(-beta.N)`
* Update the version in version.txt
* Create a new branch named `release-<version>`
* Run `make release/gen-files` to create all the manifests and bundles
* Commit the changes with `Bump version to <version>`

The steps to create the PR and the tag are still the same, and have been outlined in the README for now.


hack/release/steps.go
```diff
@@ -0,0 +1,72 @@
...
+// GenerateFiles runs make release/gen-files passing the appropriate channel options based on the version.
+func GenerateFiles(fn ExecFn) Step {
+	return StepFn(func(version string) error {
+		ch := "stable"
+		isDefault := "1"
+
+		if strings.Contains(version, "-beta") {
+			ch = "beta"
+			isDefault = "0"
+		}
```
rail, line 64: ❤️  it!

hack/release/main.go
```diff
@@ -0,0 +1,44 @@
...
+func main() {
+	flag.StringVar(&dir, "dir", ".", "the directory run in")
+	flag.StringVar(&version, "version", "", "the new version to release")
+	flag.Parse()
+
+	steps := []Step{
+		ValidateVersion(),
+		UpdateVersion(),
+		CreateReleaseBranch(process.ExecJUnit),
+		GenerateFiles(process.ExecJUnit),
```
rail, line 27: Maybe also ad a step to make sure we change the version? In other words, that the version in `version.txt` is not the same passed. I have vague memories that the `make release/gen-files` fails if you rerun it...

rail (approved):
Wooo! Great progress here.

hack/release/main.go
pseudomuto: It does indeed. Nice catch. I'll add a check

hack/release/steps.go
```diff
@@ -0,0 +1,72 @@
...
+		return ioutil.WriteFile("version.txt", []byte(version), 0644)
+	})
+}
+
+// CreateReleaseBranch creates a new branch for the release named release-<version>.
+func CreateReleaseBranch(fn ExecFn) Step {
+	return StepFn(func(version string) error {
+		return fn(
+			"git",
+			[]string{"checkout", "-b", fmt.Sprintf("release-%s", version)},
```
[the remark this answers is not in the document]
pseudomuto: I think that makes sense. Assuming `origin`_seems_ like a safe bet (I just know I'll be eating my words at some point in the future)

pseudomuto:
Updated to explicitly branch from _origin/master_ and verify that the new version isn't one of the existing git tags.

hack/release/main.go
pseudomuto: I added a check to verify the new version against the existing git tags

hack/release/steps.go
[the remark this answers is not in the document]
pseudomuto: It's in there. The trap is set 😄

hack/release/main.go
```diff
@@ -0,0 +1,64 @@
...
+
+	"sigs.k8s.io/kubetest2/pkg/process"
+)
+
+var (
+	dir     string
+	version string
+
+	deployedBundlesDir = filepath.Join("deploy", "certified-metadata-bundle", "cockroach-operator")
+	versionRegxp       = regexp.MustCompile(`^\d+\.\d+\.\d+(-beta\.\d+)?$`)
```
davidwding, line 34: nit: `deployedBundlesDir` appears unused?
nit: move `versionRegxp` to steps.go where it's used

hack/release/steps.go
```diff
@@ -0,0 +1,120 @@
...
+		}
+
+		return nil
+	})
+}
+
+// UpdateVersion sets the version in version.txt
+func UpdateVersion() Step {
+	return StepFn(func(version string) error {
+		return ioutil.WriteFile("version.txt", []byte(version), 0644)
```
davidwding, line 88: can you add a comment on what 0644 means? for those of us who aren't totally fluent in chmod values :)

hack/release/steps.go
```diff
@@ -0,0 +1,120 @@
...
+	})
+}
+
+// EnsureUniqueVersion verifies that this is a new version by checking the existing tags.
+func EnsureUniqueVersion(fn CmdFn) Step {
+	return StepFn(func(version string) error {
+		stdout := new(bytes.Buffer)
+		stderr := new(bytes.Buffer)
+
+		cmd := exec.Command("git", "tag")
```
davidwding, line 63: think it's worth using some git api like `go-git` or is that overkill?

hack/release/main.go
pseudomuto: Nice catch! Before checking the tags, I was initially going to search the deploy directory. Removed.

hack/release/steps.go
pseudomuto: Done

davidwding (approved):

PR #748, opened 2021-09-30
Leverage kustomize for generating manifests

pseudomuto:
We use two different ways to generate manifests right now. In development, kustomize bases from config are used. However in our production documentation, we ask users to apply a manifest in _manifests/operator.yaml_ which is generated with a template.

There were a number of issues I found here. Particularly, the RBAC permissions were different in the template and the generated rbac base. This was mainly due to the fact there were a lot of bindings for the cluster role defined in the template, but not as `//+kubebuilder:rbac` comments in the controller.

I've created a number of kustomizes bases here. At a high level, the three that are important for this PR are; default, manifests, and operator.

**Default**

This base contains everything you'd need to run the operator. That includes CRDs, deployments, RBAC roles/bindings, and the webhooks. This is used locally when running `make dev/up` for example.

**Manifests**

This base contains the generated CSV as well as patches for OpenShift. It includes everything from default as well as samples, scorecard and a patch to setup all of the env vars (e.g. `RELATED_IMAGE_COCKROACH_v20_1_4`) we need for deployment.

**Operator**

This is the base used to generate _install/operator.yaml_. (The root manifests directory has been renamed here to install). Similarly, install/crds.yaml is generated from the config/crds base now.

We'll need to update our docs to reflect this, but this is same atm since we recently updates the docs to point to a tagged release rather than master directly.

## Testing this out

Local development using the image from source:

```shell
$ make dev/up
# cluster should come up as expected
$ make dev/down
```

As an user following the docs using the latest (specific) release:

```shell
$ bazel run //hack:kind-start
$ kubectl apply -f install/crds.yaml
$ kubectl apply -f install/operator.yaml
$ kubectl wait --for=condition=Available deploy/cockroach-operator --timeout=2m
# cluster should come up as expected
$ bazel run //hack:kind-stop
```


manifests/operator.yaml
```diff
@@ -1,369 +0,0 @@
...
-kind: Role
-metadata:
-  name: cockroach-operator-role
-rules:
-  - apiGroups:
-      - "*"
-    resources:
-      - "*"
-    verbs:
-      - "*"
```
pseudomuto, line 72: 🔥 🔥 🔥

Makefile
```diff
@@ -289,6 +289,8 @@ release/new:
 .PHONY: release/gen-templates
 release/gen-templates:
 	bazel run //hack/crdbversions:crdbversions -- -operator-version $(APP_VERSION) -crdb-versions $(PWD)/crdb-versions.yaml -repo-root $(PWD)
+	bazel run //config/crd:manifest.preview > install/crds.yaml
```
chrislovecnm, line 292: Please add comments what we are doing

config/default/BUILD.bazel
```diff
@@ -0,0 +1,57 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+load("@k8s_deploy//:defaults.bzl", "k8s_deploy")
+load("@io_bazel_rules_k8s//k8s:objects.bzl", "k8s_objects")
+load("//hack/build:kustomize.bzl", "kustomization")
+
+kustomization(
```
chrislovecnm, line 18: More comments please. In six months we won’t remember how this is working

config/default/namespace.yaml
```diff
@@ -0,0 +1,20 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: system # will be replaced
```
chrislovecnm, line 18: Do we want the name system here? is-replaced or something like that

config/manager/deployment.yaml
```diff
@@ -0,0 +1,53 @@
...
+  template:
+    metadata:
+      labels:
+        app: cockroach-operator
+    spec:
+      serviceAccountName: cockroach-operator-sa
+      containers:
+        - name: cockroach-operator
+          image: cockroachdb/cockroach-operator:latest
+          imagePullPolicy: IfNotPresent
```
chrislovecnm, line 35: Side note, we may want to change this.

config/manager/kustomization.yaml
```diff
@@ -11,11 +11,9 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
-namespace: default
-
+---
 resources:
-- operator.yaml
```
chrislovecnm, line 18: The indentation is different. Is this deliberate?

pkg/controller/cluster_controller.go
```diff
@@ -58,9 +58,10 @@ type ClusterReconciler struct {
 // +kubebuilder:rbac:groups=core,resources=configmaps,verbs=get
 // +kubebuilder:rbac:groups=core,resources=configmaps/status,verbs=get
 // +kubebuilder:rbac:groups=core,resources=secrets,verbs=get;list;watch;create;update;patch
+// +kubebuilder:rbac:groups=core,resources=pods,verbs=get;list
```
chrislovecnm, line 61: Did you take a look at all of the RBAC rules? We probably should.

chrislovecnm (commented):
Love love love this PR. Been meaning to do this for awhile. Just some comments. Nothing big.

config/manager/deployment.yaml
pseudomuto: I'm assuming you mean the image (and not the policy). The `:latest` tag here is just a placeholder. All of our codegen overrides this with the specific latest release. See _config/manager/patches/image.yaml_ which is generated from a template.

pkg/controller/cluster_controller.go
pseudomuto: I started by removing everything that was in operator.yaml.in and then adding back the minimal set needed to get the operator to run correctly. I didn't try removing any of the existing markers. I think we should, but I'll tackle that in a separate PR.

config/manager/kustomization.yaml
pseudomuto: Must have got formatted when I added the `patchesStrategicMerge`

Makefile
pseudomuto: Added some more details to the make target comment, as well as on the kustomization macros in the build files

chrislovecnm (approved):
Looks awesome. One comment is, how do we unit test this? Please file an issue and I am merging.

PR #750, opened 2021-10-01
Make it so that at most one actor runs per reconcile loop

davidwding:
This PR makes it so that at most one actor runs per reconcile loop. If no actors need to run, nothing happens.

This has several benefits:
- The operator becomes much more like a state machine. Rather than blindly looping through many actors with every reconcile (and polluting the logs with tons of irrelevant information), the operator runs only what it needs to.
- All the business logic that determines when an actor needs to run is in one place, as opposed to scattered across all the actors, making debugging easier.
- When each reconcile loops through many actors, it's possible that actors step on each others' toes and cause various issues. For example, suppose somebody changes the spec to trigger a decommission right before a reconcile loop reaches the deploy step. Then deploy may actually run first and deactivate a node before it actually gets decommissioned.

The meat of this PR centers around the new `GetActorToExecute` func, which calls helper funcs to determine if an actor needs to run. This is not too complex for the most part, with the exception of the `Deploy` actor, which required shuffling some deployment logic around to make it easy to determine if some component of the operator has changed.

I also added a lot of unit tests in director_test.go.


davidwding (commented):
@pseudomuto just as an FYI, I've integrated the new setup RBAC actor into this PR, added testing for it, and verified that it still works fine on a GKE cluster.

pseudomuto (approved):

PR #756, opened 2021-10-05
Update CertificatesV1Beta1 to CertficatesV1

### LINKED ISSUE #662: pkg/tls/csr.go is still using the certificates.k8s.io/v1beta1 for something, which is removed in k8s 1.22
Either this code isn't being used anymore and should be removed, or it needs to be updated to the v1 api.  If its being used this will block anyone on 1.22 from using our operator.

pseudomuto:
Fixes #662 

Kubernetes 1.22 has dropped the v1beta version of the Certificates API. See <https://kubernetes.io/blog/2021/07/14/upcoming-changes-in-kubernetes-1-22/#api-changes> for more information.

Fortunately for us, we're not actually using the code we have here that references the v1beta1 API. So I've removed it here.

In preparation for OpenShift 4.9 (which uses k8s 1.22) I've verified that were' not using any of the removed APIs from the list above.

**Checklist**

* [ ] I have added these changes to the changlog (or it's not applicable).


hack/dev.sh
```diff
@@ -34,6 +35,9 @@ create_kind_cluster() {
   cat <<EOF | kind create cluster --name "${CLUSTER_NAME}" --config=-
 apiVersion: kind.x-k8s.io/v1alpha4
 kind: Cluster
+nodes:
+- role: control-plane
+  image: ${NODE_IMAGE}
```
pseudomuto, line 40: This ensures we run 1.22.1 in dev

udnay:
We also have https://github.com/cockroachdb/cockroach-operator/blob/master/bundle/manifests/crdb.cockroachlabs.com_crdbclusters.yaml#L14 which was removed. That needs to go in here too, right?

pseudomuto:
As we discussed in Slack:

> We also have https://github.com/cockroachdb/cockroach-operator/blob/master/bundle/manifests/crdb.cockroachlabs.com_crdbclusters.yaml#L14 which was removed. That needs to go in here too, right?

That bundle file is misleading. We don't actually use it. We use the files in deploy/certified-metadata-bundle, which all reference v1.

> does this bump the CRD to ga? The api version on the crd. I thought we needed to bump that

We need to ensure our CRD is defined with the v1 API for CRDs (it is). We can remain at v1alph1

PR #759, opened 2021-10-07
Add nightly tests for supported K8s versions

pseudomuto:
We need to ensure that we're testing all of the supported versions of Kubernetes. As APIs are upgraded/removed we should not rely on manually verifying which APIs are being used and which are still valid.

I've added a smoke test here which stands up a kind cluster (for each supported version), deploys the operator, and runs the bank workload for a few seconds just to ensure things are working.

You might be thinking, "why not just run (one of) our e2e tests?" The answer is that I wanted to ensure we had a test in place that does what our [docs] tell users to do. Which is to install the crds from _install/crds.yaml_, the operator from _install/operator.yaml_, and the example from _examples/example.yaml_. These files are updated whenever we do a release or run `make release/gen-templates`.

To make sure the cluster is functional, the test also adds the secure client pod and executes the [bank workload] for 10s.

Finally, I've added a new GitHub workflow to run these tests nightly. I chose 3am rather arbitrarily since I figured almost everyone else chooses midnight :smile:

[docs]:
https://www.cockroachlabs.com/docs/stable/deploy-cockroachdb-with-kubernetes.html#step-2-start-cockroachdb
[bank workload]:
https://www.cockroachlabs.com/docs/stable/cockroach-workload.html#run-the-bank-workload

**Checklist**

### DIFF
--- a/.github/workflows/nightly-smoketest.yaml
+++ b/.github/workflows/nightly-smoketest.yaml
@@ -1,0 +1,46 @@
+# Copyright 2021 The Cockroach Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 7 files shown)


davidwding (approved):
LGTM but let's make the steps stuff a bit more DRY in a later PR

PR #760, opened 2021-10-07
Removing bundle directory and update-csv.sh

### LINKED ISSUE #225: Openshift Documentation for CSV generation
- Document update-csv.sh that runs from bazel

pseudomuto:
Fixes #225 

We don't actually use this directory. We use bundles in _deploy/certified-metadata-bundle_ instead. This has caused some confusion in terms of which things need to be updated during a deploy for example.

So I've removed them and the associated script and make target.

**Checklist**


pseudomuto:
Indeed...used `make test/openshift-package VERSION=2.2.1` to verify

davidwding (dismissed):

PR #761, opened 2021-10-07
Check desired replicas vs ready replicas in e2e tests

### LINKED ISSUE #655: e2e tests are not waiting for the pod status set as ready
We check that the pod is running, but not that the pod is ready. We terminate the e2e tests when we have:

```
crdb-0   0/1     Running   0          10s
```

And the status should be Ready.

pseudomuto:
Fixes #655 

I was looking to close #655 and found that it has mostly been done already. I did notice we were comparing status to status, rather than spec to status. This is just small fix to check the spec.Replicas vs Status.ReadyReplicas on the StatefulSet.

**Checklist**

### DIFF
--- a/pkg/testutil/require.go
+++ b/pkg/testutil/require.go
@@ -22,14 +22,11 @@ import (
 	"encoding/csv"
 	"fmt"
 	"io"
-	v1 "k8s.io/api/batch/v1"
 	"os"
 	"strconv"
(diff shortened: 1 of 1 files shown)


chrislovecnm (commented):
Looks good but decommission e2e failed.  I filed an issue.

PR #764, opened 2021-10-07
Update README.md

udnay:
PR #748 changed where our manifests are, this PR updates the README with the new locations.

**Checklist**

* [x] I have added these changes to the changlog (or it's not applicable).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -61,6 +61,5 @@ bazel-testlogs
 .secrets
 opm
 faq
-jobs/*
 
 _artifacts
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -61,6 +61,5 @@ bazel-testlogs
 .secrets
 opm
 faq
-jobs/*
```
pseudomuto, line 64: This was done simply to force the required CI checks to run. We don't have a jobs directory, so should be safe.

PR #767, opened 2021-10-08
Always run required checks

pseudomuto:
We recently ran into an issue where an update to README wouldn't run the required checks. It would make sense to skip the checks for docs changes, but the checks are required for merging. Leading us to change .gitignore simply to trigger them.

We can look into a better long-term solution, but for now I've altered the workflows to run on every push/pr to master.

**Checklist**


udnay (approved):
:shipit:

PR #776, opened 2021-10-12
Some restructuring within Actor and Director

davidwding:
This PR contains no changes to business logic. It consists of miscellaneous structural changes that are prerequisites for the big refactor PR, which I moved into this PR to keep review manageable.

Specific changes:
- Move director stuff out of actor.go into director.go, and rename actor_test.go to director_test.go.
- Clean up the `action` struct and associated structs a bit: `config` was in every struct that embedded `action` so I just moved the field into `action`.
- Introduce `clientset` as a field provided to all the actors, through the director, rather than constructing it anew in each actor. This makes it possible to do some mocking which is necessary for proper unit testing. 
- Remove uses of `NewFakeClientWithScheme`, which is deprecated, and replace with the recommended func.

**Checklist**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 # [Unreleased](https://github.com/cockroachdb/cockroach-operator/compare/v2.2.1...master)
 
+## Added
+
+* Support for nodeSelectors
+
 ## Changed
 
 * Cleanup of initialization and structure of Actor and Director
--- a/apis/v1alpha1/cluster_types.go
+++ b/apis/v1alpha1/cluster_types.go
@@ -118,6 +118,10 @@ type CrdbClusterSpec struct {
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Cockroach Database Tolerations"
 	// +optional
 	Tolerations []corev1.Toleration `json:"tolerations,omitempty"`
+	// (Optional) If specified, the pod's nodeSelector
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Map of nodeSelectors to match when scheduling pods on nodes"
+	// +optional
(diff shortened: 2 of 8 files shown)


pkg/actor/actor.go
```diff
@@ -79,95 +74,19 @@ type Actor interface {
...
 }
 
 // action is the base set of common parameters required by other actions
 type action struct {
-	client client.Client
-	scheme *runtime.Scheme
+	client    client.Client
+	clientset kubernetes.Interface
+	scheme    *runtime.Scheme
+	config    *rest.Config
```
pseudomuto, line 91: If we've got a clientset, do we need scheme and/or config?

pkg/actor/deploy_test.go
```diff
@@ -73,7 +73,7 @@ func TestDeploysNotInitalizedClusterAfterVersionChecker(t *testing.T) {
 	cluster.SetTrue(api.CrdbVersionChecked)
 
 	mock := kube.MockKubernetesDistribution()
-	deploy := actor.NewDeploy(scheme, client, nil, mock)
+	deploy := actor.NewDeploy(scheme, client, nil, mock, nil)
```
pseudomuto, line 76: rather than passing `nil` for unused values could we update the signatures? We require that these functions return `Actor` instances, but we don't need to satisfy any other interface right?

pseudomuto (commented):
I'm digging the idea of reusing a pre-initialized kubernetes.Interface 👍

pkg/actor/actor.go
davidwding: Unfortunately they do seem to be used sporadically throughout the Actors...maybe it's possible to refactor them away but I think that should be done in a later PR (if at all)?

pkg/actor/deploy_test.go
davidwding: I did a (relatively naive) audit of all the signatures of the actor constructors and cleaned up a lot. There are still some instances in the tests where pass a `nil` but the actor requires it, and we're just not testing some specific flow.

PR #785, opened 2021-10-19
Added ingress support #712

### LINKED ISSUE #712: Expose CRDB services via Ingress
There should be a way for the user to expose CRDB services to the internet via Ingress

abhishekdwivedi3060:
closes #712 

- Added support for UI ingress resource with CRDB cluster
- Added new actor for ingress resource creation and deletion


apis/v1alpha1/webhook.go
```diff
@@ -82,13 +84,34 @@ func (r *CrdbCluster) Default() {
 // ValidateCreate implements webhook.Validator so a webhook will be registered for the type.
 func (r *CrdbCluster) ValidateCreate() error {
 	webhookLog.Info("validate create", "name", r.Name)
+	var errors []error
```
davidwding, line 87: why do we need to have a slice of errors if there can be at most one error?

pkg/actor/director.go
```diff
@@ -352,3 +360,50 @@ func (cd *clusterDirector) needsInitialization(cluster *resource.Cluster) bool {
...
+
+	// In order to expose ingress,
+	// - the cluster initialized condition must be true
+	// - if there is a change in ingress resource
+
+	if !conditionInitializedTrue {
+		return false, nil
+	}
+
+	// this is the case of update when ingress is removed from CR
```
davidwding, line 377: nit: maybe name the function `needsIngressDeploy`? it's a little confusing in this case when the name is `needsIngress` but we need to remove the ingress

pkg/controller/cluster_controller.go
```diff
@@ -210,12 +216,19 @@ func (r *ClusterReconciler) Reconcile(ctx context.Context, req reconcile.Request
 
 // SetupWithManager registers the controller with the controller.Manager from controller-runtime
 func (r *ClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	var ingress client.Object
+	ingress = &v1.Ingress{}
+	if !util.CheckIfAPIVersionKindAvailable(mgr.GetConfig(), "networking.k8s.io/v1", "Ingress") {
```
davidwding, line 221: why do we need this - does it depend on the version of k8s being used or something?

apis/v1alpha1/webhook.go
abhishekdwivedi3060: Right now, there are minimal checks in webhook so returning single error is fine.
But as we keep on adding more validation checks in webhook, ideally we should validate the complete CR object and return the combined error in one go instead of returning one by one error in each iteration by the user.
Hence added slice of errors here.

davidwding (approved):

PR #793, opened 2021-10-21
Update smoke tests

pseudomuto:
With v2.3.0 release, we changed how webhook secrets are used. So I've updated the smoketest script accordingly.

I've also added the ability to override some values in the manifests applied in the smoketest. Right now they don't run because the pods can't be scheduled due to the memory requirements. We may need to play around with this but for now I've just cut the requirements in half when running these.

**Checklist**


hack/smoketest/main.go
```diff
@@ -64,8 +64,15 @@ func main() {
...
 		WaitForDeploymentAvailable("cockroach-operator"),
-		WaitForSecret("cockroach-operator-webhook-tls"),
-		ApplyManifest(filepath.Join("examples", "example.yaml")),
+		WaitForSecret("cockroach-operator-webhook-ca"),
+		sleep(10 * time.Second), // just give the manager time to write the TLS certs to disk
+		ApplyManifestWithReplacements(
+			filepath.Join("examples", "example.yaml"),
+			map[string]string{
+				// 2Gi was preventing the pods from getting scheduled in GitHub actions
+				"memory: 2Gi": "memory: 1Gi",
```
davidwding, line 73: should we just update example.yaml to be 1Gi? i feel i always replace it with 1Gi manually too

PR #795, opened 2021-10-21
Deprecate OpenShift package manifest format in favor of bundle format

davidwding:
As noted [here](https://docs.openshift.com/container-platform/4.8/operators/understanding/olm-packaging-format.html), the `packagemanifest` packaging format is being deprecated in favor of the `bundle` packaging format.

We are currently doing a two-step process where we generate files for a new OpenShift release using the `packagemanifest` command in `update-pkg-manifest.sh`, then converting it to the `bundle` format in `opm-build-bundle.sh`. This PR makes it so that we directly generate the right files with `operator-sdk` using a single updated command, and then do some processing so that files are ultimately laid out in the same way that they currently are.

I tested this out on an OpenShift cluster I spun up on GCP and verified that the operator works. Check out [this sample PR](https://github.com/cockroachdb/cockroach-operator/pull/794) for an example of what this updated release script generates.

**Checklist**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 * Deprecated legacy OpenShift packaging format in favor of new bundle format
 
+## Fixed
+
+* Boilerplate test after updating to Go 1.17
+
 # [v2.3.0](https://github.com/cockroachdb/cockroach-operator/compare/v2.2.1...v2.3.0)
 
 ## Added


hack/update-pkg-bundle.sh
```diff
@@ -53,18 +53,21 @@ ensure_unique_deployment() {
...
-generate_package_manifest() {
+generate_package_bundle() {
   # Generate CSV in config/manifests and add boilerplate back (lost when regenerating)
   operator-sdk generate kustomize manifests -q --apis-dir apis
   hack/boilerplaterize hack/boilerplate/boilerplate.yaml.txt config/manifests/**/*.yaml
 
-  # TODO: packagemanifests is deprecated and scheduled to be remove in 2.0.0
-  kustomize build config/manifests | operator-sdk generate packagemanifests -q \
+  # Generate the new package bundle
+  kustomize build config/manifests | operator-sdk generate bundle -q \
```
davidwding, line 62: One thing I want to explicitly call out is that one of the annotations changes from cockroachdb-certified to cockroach-operator, as you can see here: https://github.com/cockroachdb/cockroach-operator/pull/794/files#r733931636

cockroachdb-certified doesn't appear to be anywhere in our codebase but has been automatically propagated from the previous version with every new release. Will this cause any issues?

PR #811, opened 2021-10-29
Make CRDBVersion and Image section optional in crdbcluster 

prafull01:
Currently the image field is mandatory to provide and even if user has provided the `cockroachDBVersion`, he has to provide the image which is not used at all.

This PR makes both the field optional and validates in admission webhooks that user must provide one output to be able to go forward
**Checklist**


apis/v1alpha1/cluster_types.go
```diff
@@ -37,8 +37,8 @@ type CrdbClusterSpec struct {
 	Nodes int32 `json:"nodes"`
 	// Container image information
```
pseudomuto, line 38: Seems like we've got a convention here to prefix these comments with `(Optional) ` for optional fields

apis/v1alpha1/webhook.go
```diff
@@ -83,6 +84,9 @@ func (r *CrdbCluster) Default() {
 func (r *CrdbCluster) ValidateCreate() error {
 	webhookLog.Info("validate create", "name", r.Name)
 
+	if r.Spec.CockroachDBVersion == "" && (r.Spec.Image == nil || r.Spec.Image.Name == "") {
```
pseudomuto, line 87: Should we do the same check for updates?

pseudomuto: Also, I think we're missing a test case for this

apis/v1alpha1/cluster_types.go
prafull01: Done

apis/v1alpha1/webhook.go
prafull01: Added for updates as well and added another check if both is provided by the user then also give warning. Also, added the tests for the same

davidwding: these checks look a little complex and are duplicated across `ValidateCreate` and `ValidateUpdate` - can you put them into a helper func?

prafull01: Done

apis/v1alpha1/webhook_test.go
```diff
@@ -68,3 +72,62 @@ func TestValidateIngress(t *testing.T) {
...
+		Spec: CrdbClusterSpec{
+			Image: &PodImage{Name: "testImage"},
+			CockroachDBVersion: "v2.1.20",
+		},
+	}
+
+	err := cluster.ValidateUpdate(&oldCluster)
+	require.Error(t, err)
+	require.Equal(t, err.Error(), "you have provided both cockroachDBVersion and cockroach image, please provide only one")
+}
```
pseudomuto, line 133: nit: missing new line

pseudomuto (approved):
LGTM 👍 

I think we could make those tests table tests to avoid some duplication, but no biggie.

apis/v1alpha1/webhook_test.go
prafull01: Done, added table driven tests as well

PR #812, opened 2021-11-01
Dynamically create database-sa for CrdbCluster resources

pseudomuto:
Currently we predefine the service account, role, and role binding for CrdbCluster instances. As a first step towards making the operator work in multiple namespaces (and with multiple clusters) we need to create these resources per instance.

I've added a new actor (setupRBACAction) which creates a new namespaced SA for the cluster, along with a namespaced role, and role-binding. These are owned by the cluster and are deleted when the cluster resource is deleted.

When a new CR is created, the SA will be created (if necessary), the role will be created (if necessary), and the SA will be added to the role bindings subjects (again if necessary).

There's nothing in this PR to make the operator work with multiple namespaces, but we should be able to run multiple CRDB clusters in the same namespace, each with their own SA.

**Checklist**

# Testing

Start by creating a GKE cluster:

```bash
bazel run //hack/gke -- \
  -action start \
  -cluster-name multi-crs \
  -gcp-project <GCP_PROJECT> \
  -gcp-zone <GCP_ZONE>
```

Run the following script to publish the image to GCR and deploy the operator in your new cluster:

```bash
#!/usr/bin/env bash
set -euo pipefail

export REGISTRY="gcr.io/<GCP_PROJECT>"
export PATH="bazel-bin/hack/bin:${PATH}"

main() {
  generate_code
  publish_image
  install_crds
  install_operator
  kubectl config set-context --current --namespace crdb
}

generate_code() {
  make dev/generate release/gen-templates
}

publish_image() {
  DOCKER_REGISTRY="${REGISTRY}" \
    DOCKER_IMAGE_REPOSITORY="cockroachdb/cockroach-operator" \
    make release/image
}

install_crds() {
  kubectl apply -f install/crds.yaml
}

install_operator() {
  sed \
    -e "s+image: +image: ${REGISTRY}/+g" \
    -e "s/: default/: crdb/" \
    -e "s/: IfNotPresent/: Always/" \
    install/operator.yaml | kubectl apply -f -
}

main "$@"
```

Deploy the example CR and verify roles:

```bash
kubectl apply -f examples/example.yaml
kubectl get sa # should see cockroachdb-sa
kubectl get roles # should see cockroachdb-role
kubectl describe rolebindings cockroachdb-role-binding # should see SA in the subjects field
```

Deploy another cluster resource:

```bash
sed '1,/cockroachdb/s/cockroachdb/crdb/' examples/example.yaml | kubectl apply -f -
kubectl get sa # should see crdb-sa
kubectl get roles # should see crdb-role
kubectl describe rolebindings crdb-role-binding # should see SA in the subjects field
```

There should be no issues reconciling the instances.

Delete a CrdbCluster instance:

```bash
kubectl delete -f examples/example.yaml
kubectl get sa # should NOT see cockroachdb-sa
kubectl get roles # should NOT see cockroachdb-role
kubectl get rolebindings # should NOT see cockroachdb-role-binding
```

Cleanup:

```bash
bazel build //hack/bin:kustomize # ensure kustomize is available
PATH="./bazel-bin/hack/bin:${PATH}" hack/delete-gke-cluster.sh -c multi-crs
```


apis/v1alpha1/action_types.go
```diff
@@ -19,27 +19,18 @@ package v1alpha1
...
-	DecommissionAction ActionType = "Decommission"
-	//InitializeAction string
-	InitializeAction ActionType = "Initialize"
-	//RequestCertAction string
-	RequestCertAction ActionType = "RequestCert"
-	//GenerateCert string
-	GenerateCertAction ActionType = "GenerateCert"
-	//RequestCertAction string
-	ResizePVCAction ActionType = "ResizePVC"
-	//UpgradeAction string
```
pseudomuto, line 39: These comments didn't seem valuable

pkg/controller/cluster_controller.go
```diff
@@ -75,6 +78,7 @@ type ClusterReconciler struct {
 // +kubebuilder:rbac:groups=batch,resources=jobs/finalizers,verbs=get;list;watch
 // +kubebuilder:rbac:groups=admissionregistration.k8s.io,resources=mutatingwebhookconfigurations,verbs=get;update;patch;
 // +kubebuilder:rbac:groups=admissionregistration.k8s.io,resources=validatingwebhookconfigurations,verbs=get;update;patch;
+// +kubebuilder:rbac:groups=security.openshift.io,resources=securitycontextconstraints,verbs=use
```
pseudomuto, line 81: This is needed because the operator SA can't grant permissions it doesn't have. Since we grant `use security.openshift.io/securitycontextconstraints` to the database role, we need it here as well.

udnay (commented):
The code looks good, do we need to update the examples now as well? Specifically the secure-client yaml

pseudomuto:
Updated the client-secure-pod (and template). By default, this will work for the example cluster resource (using the correct service account). I verified this by installing the pod and running a cockroach sql command from it.

pkg/actor/director.go
```diff
@@ -78,6 +87,13 @@ func (cd *clusterDirector) GetActorsToExecute(cluster *resource.Cluster) []Actor
 
 	var actorsToExecute []Actor
 
+	// we need to ensure the SA exists early on since the version checker job uses it which happens before the deploy.
+	ctx := context.Background()
+	serviceAccounts := cd.clientset.CoreV1().ServiceAccounts(cluster.Namespace())
+	if _, err := serviceAccounts.Get(ctx, cluster.ServiceAccountName(), metav1.GetOptions{}); kube.IsNotFound(err) {
```
davidwding, line 93: sweet, this should make it easy to merge in with the one-actor PR (whichever comes first)

davidwding (approved):

PR #813, opened 2021-11-01
Add sts/scale permissions for operator role

pseudomuto:
When testing out  an upcomfing change to how we determine which actor to run, we noticed there were some error messages when scaling down (decommissioning) a cluster. The issue is that we aren't able to update the scale subresource for statefulsets.

I've added the permissions and verified that it works locally and in a GKE cluster.

**Checklist**


davidwding (approved):
🔥

PR #819, opened 2021-11-10
Update bundle generation code

pseudomuto:
There's been a change to the way OpenShift bundles are certified which changes our release process. I've updated the code here to drop the bundles into the following directories, which represent the certified bundle and the marketplace bundle respectively.

* `bundle/<version>/cockroachdb-certified`
* `bundle/<version>/cockroachdb-certified-rhmp`

After we cut a release (via `make release/new`), we'll need to perform the following steps:

* Publish docker image (no changes)
* Tag the release (no changes)
* Upstream PR with the certified bundle (e.g.
  redhat-openshift-ecosystem/certified-operators#81)
* Upstream PR with the marketplace bundle (e.g.
  redhat-openshift-ecosystem/redhat-marketplace-operators#5)

**Checklist**


bundle.Dockerfile
```diff
@@ -4,7 +4,7 @@ FROM scratch
 LABEL operators.operatorframework.io.bundle.mediatype.v1=registry+v1
 LABEL operators.operatorframework.io.bundle.manifests.v1=manifests/
 LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
-LABEL operators.operatorframework.io.bundle.package.v1=cockroachdb-certified
+LABEL operators.operatorframework.io.bundle.package.v1=cockroach-operator
```
pseudomuto, line 7: Nothing to worry about here, this file is auto-generated but not used. We could gitignore it if that's preferred.

bundle/2.4.0/cockroachdb-certified-rhmp/manifests/cockroach-operator-webhook-service_v1_service.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: v1
```
davidwding, line 1: We used to "combine" these all into one yaml file - probably doesn't matter if we don't?

davidwding (approved):

bundle/2.4.0/cockroachdb-certified-rhmp/manifests/cockroach-operator-webhook-service_v1_service.yaml
pseudomuto: Correct. This isn't needed so I opted for simple

PR #826, opened 2021-11-26
Update smoke tests to be aware of new namespace

pseudomuto:
Since updating the default namespace to `cockroach-operator-system` the nightly smoke tests have been failing. This is because there is a check to wait for a deployment (that doesn't exist) to be available.

This patch adds the namespace to the appropriate calls and makes the tests work again. I've verified this by running the following command locally: `make test/smoketest NODE_VERSION=1.22.1`

**Checklist**


abhishekdwivedi3060 (commented):
LGTM

abhishekdwivedi3060 (approved):

Issue #613, opened 2021-06-21
operator incorrectly uses `--logtostderr` + can't use new --log flag with Operator


taroface:
I attempted to specify the new [`--log` flag](https://www.cockroachlabs.com/docs/v21.1/configure-logs#flag) in the custom resource:

```
  additionalArgs:
    - "--log='sinks: {file-groups: {ops: {channels: [OPS, HEALTH]}}}'"
```

This resulted in the pods crashing:

```
NAME                                 READY   STATUS             RESTARTS   AGE
cockroach-operator-d4df75996-pdpmd   1/1     Running            0          135m
cockroachdb-0                        0/1     CrashLoopBackOff   11         34m
cockroachdb-1                        0/1     CrashLoopBackOff   11         34m
cockroachdb-2                        0/1     CrashLoopBackOff   11         34m
```

The kubectl logs show:

```
++ expr 8192 / 4
++ expr 8192 / 4
+ exec /cockroach/cockroach.sh start --join=cockroachdb-0.cockroachdb.default:26258,cockroachdb-1.cockroachdb.default:26258,cockroachdb-2.cockroachdb.default:26258 --advertise-host=cockroachdb-0.cockroachdb.default --logtostderr=INFO --certs-dir=/cockroach/cockroach-certs/ --http-port=8080 --sql-addr=:26257 --listen-addr=:26258 --cache 2048MiB --max-sql-memory 2048MiB '--log=sinks: {file-groups: {ops: {channels: [OPS, HEALTH]}}}'
Flag --logtostderr has been deprecated, use --log instead to specify 'sinks: {stderr: {filter: ...}}'.
E210621 20:27:35.192687 1 1@cli/error.go:399  [-] 1  ERROR: --log is incompatible with legacy discrete logging flags
ERROR: --log is incompatible with legacy discrete logging flags
```

Since `--logtostderr` is deprecated in v21.1 (and according to @knz we shouldn't be logging to stderr anyway), perhaps this [line](https://github.com/cockroachdb/cockroach-operator/blob/master/pkg/resource/statefulset.go#L340) in the code should be removed. I believe the logging behavior would then default to [this configuration](https://www.cockroachlabs.com/docs/v21.1/configure-logs#default-logging-configuration).

If we do need to log to `stderr` on Kubernetes, I believe the equivalent v21.1 flag is:

```
--log="sinks: {stderr: {channels: all, filter: INFO}}"
```

[...]

PR #827, opened 2021-11-29
Add the support of custom logging configuration for cockroach version greater than v21.1

prafull01:
Added the logging configuration feature for cockroachdb as per the following design [document](https://docs.google.com/document/d/1bcn7FRPH4COwaiJakZrz6pdYk7e8RtrMwHL-SJsk5EM/edit)

Highlights of the changes:
- CRD changes for taking the logging configmap input for cockroachdb.
- The logging configmap should have the logging configs in form of json/yaml. See the following example:
```
apiVersion: v1
data:
  logconfig.json: |
    {
      "sinks": {
        "file-groups": {
          "dev": {
            "channels": "DEV",
            "filter": "WARNING"
          }
        }
      }
    }
kind: ConfigMap
metadata:
  name: logconfig
  namespace: default
```
Or
```
apiVersion: v1
data:
  logconfig.yaml: |
    sinks:
      file-groups:
        dev:
          channels: DEV
          filter: WARNING
kind: ConfigMap
metadata:
  name: logcm
  namespace: default
```
- Users provide the configmap name while creating the `CrdbCluster` CR and provide it in `spec` section as follows:
```
spec:
  logConfigMap: logconfig
```

- For cockroach `version >= v21.1.0`, the logging API is passed to cockroachdb start. If user doesn't provide logging configmap, by default the logging config is:
```
sinks:
    stderr:
        channels: OPS
        redact: true
```
- For cockroach `version <= v20.x`, the logs are sent to stderr `--logtostderr=INFO`
- Before starting the cluster, it validates the logging API input is proper or not with `cockroach debug check-log-config --log=<log yaml>`, if the logging input is proper then only we create the cluster.
- Upgraded the cockroach binary version from `v20.2.5` to `v21.1.0` inside the cockroach operator pod.

**Checklist**


pseudomuto (commented):
Overall looking good 👍 I like that we're just using crdb to validate the configuration

pkg/resource/cluster.go
```diff
@@ -295,6 +298,47 @@ func (cluster Cluster) IsFresh(fetcher Fetcher) (bool, error) {
...
 
+func (cluster Cluster) LoggingConfiguration() (string, error) {
+	var logYaml string
+	if cluster.Spec().LogConfig != nil {
+		logYamlBytes, err := json.Marshal(cluster.Spec().LogConfig.LogFile)
+		if err != nil {
+			return "", err
+		}
+		var out bytes.Buffer
+		err = json.Indent(&out, logYamlBytes, "", " ")
```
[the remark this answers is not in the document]
prafull01: It doesn't matter as the this yaml is supplied as a string to the `cockroach start` command as --log=<formatted_yaml_string>.

apis/v1alpha1/cluster_types.go
```diff
@@ -127,6 +128,11 @@ type CrdbClusterSpec struct {
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Cockroach Database Ingress"
 	// +optional
 	Ingress *IngressConfig `json:"ingress,omitempty"`
+	// (Optional) LogConfig define the log configuration used to send the logs through the proper channels in the
```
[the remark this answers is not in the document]
prafull01: Done

pkg/resource/cluster.go
```diff
@@ -295,6 +298,47 @@ func (cluster Cluster) IsFresh(fetcher Fetcher) (bool, error) {
...
+		}
+		logYaml = "\"" + out.String() + "\""
+	} else {
+		logYaml = "\"{sinks: {stderr: {channels: OPS, redact: true}}}\""
+	}
+
+	return logYaml, nil
+}
+
+func (cluster Cluster) IsLoggingAPIEnable() bool {
```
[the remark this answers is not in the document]
prafull01: Done

hack/bin/deps.bzl
```diff
@@ -432,13 +432,13 @@ filegroup(
 def install_crdb():
     http_archive(
        name = "crdb_darwin", # todo fix or remove
-       sha256 = "bbe3a03c661555e8b083856c56c8a3b459f83064d1d552ed3467cbfb66e76db7",
-       urls = ["https://binaries.cockroachdb.com/cockroach-v20.2.5.darwin-10.9-amd64.tgz"],
+       sha256 = "292c9536ce8526ab71befd18431aa8e1da3f5a83ca92af8bfbde501254bdb148",
+       urls = ["https://binaries.cockroachdb.com/cockroach-v21.1.0.darwin-10.9-amd64.tgz"],
```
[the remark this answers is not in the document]
prafull01: Done

pkg/resource/testdata/TestStatefulSetBuilder/default_insecure_statefulset.golden
```diff
@@ -46,7 +46,7 @@ spec:
             resourceFieldRef:
               divisor: 1Mi
               resource: limits.memory
-        image: cockroachdb/cockroach:v20.2.7
+        image: cockroachdb/cockroach:v21.1.0
```
pseudomuto, line 49: Should be v21.2.0?

pseudomuto (approved):

pkg/resource/testdata/TestStatefulSetBuilder/default_insecure_statefulset.golden
prafull01: No, we have upgraded the cockroach binary to `v21.2.0` in operator pod, however logging APIs are supported from `v21.1.0`, hence in test cases I have added the minimum supported version. This way we can test the statefulset yaml is properly generated on this minimum version

pkg/resource/testdata/TestStatefulSetBuilder/default_insecure_statefulset_in.yaml
```diff
@@ -31,7 +31,7 @@ spec:
   grpcPort: 26258
   httpPort: 8080
   image:
-    name: cockroachdb/cockroach:v20.2.7
+    name: cockroachdb/cockroach:v21.1.0
```
udnay, line 34: We should make sure that this is `v21.1.1` or higher

udnay (approved):
LGTM

pkg/resource/testdata/TestStatefulSetBuilder/default_insecure_statefulset_in.yaml
prafull01: The design document suggest the logging api is supported `21.1.x` version and the lowest version is `v21.1.0`. Is there any specific reason we want to exclude `v21.1.0` and start supporting from `v21.1.1`?

udnay:
@knz is there a way to ensure that all logs go to stderr or a specific sink?

knz:
> @knz is there a way to ensure that all logs go to stderr or a specific sink?

Can you clarify this question? What do you mean with "all logs", and what do you mean with "or" in that sentence? Can you give an example?

udnay:
@prafull01 was saying that with the new settings there are some logs that are still written to disk even with the provided yaml stanza of
```yaml
sinks:
    stderr:
        channels: OPS
        redact: true
```
Is there a way to have all channels write to the `stderr` sink? This would be the default if the user does not specify any logging configuration and would preserve the current behaviour.

knz:
> @prafull01 was saying that with the new settings there are some logs that are still written to disk

Okay let's take a deep breath and a step back, because it looks like the participants in this convo have not yet familiarized themselves with the issues described in #613.

I'll repeat what I [explained before](https://github.com/cockroachdb/cockroach-operator/issues/613#issuecomment-870761851):

> It's good to see the _default_ stderr output (which is just a server startup message) in the output of `kubectl log`. we might have some interest in seeing the output of the OPS logging channel (new in 21.1) via `kubectl log` but I don't know that we have evidence for that (yet).
> Whatever additional detailed logging produced by crdb **should not be available via `kubectl log`** (emphasis added 2021-12-13). The logging is largely structured and inadequate for channeling through kube's simplistic logging output. We should instead require an operator to run `cockroach debug zip` or some other related command to extract operational details.

Then I invite the participants to re-familiarize themselves with [this comment](https://github.com/cockroachdb/cockroach-operator/issues/613#issuecomment-870883571):

- (point 2) logging to stderr creates output that cannot be reliably parsed.  Presumably, customers who integrate with orchestration wish the ability to process logging data automatically. The output produced by crdb on its stderr stream fails at this. Even though it looks like something parseable on the surface, it is a documented fact that _under stress and in problem situations, the log entries become malformed_. This is not fixable via stderr logging, but is fixable via file or network logging.

  Right now the default config used by the k8s operator code is mistakenly giving users the impression they can rely on the logging output by `kubectl log` and automate log monitoring etc based on that data. This is a lie to our users/customers.

- (point 3) you and others who are working on the operator project are working under the mistaken impression that "logging" is somehow distinct from "database state". I see this for example in your suggestion that _"if we store log files we'd need ephemeral storage"_ or the question _"how would the operator look at the logs without execing into the node"_. That is something we need to work on culturally. 

  In a nutshell, crdb logging data has the same quality as general stored SQL data.

  We don't think about "how would the operator access the SQL data without execing into the node", right? To access SQL data, we open a SQL client. 

  The same applies to crdb logs. To access the log data, we use the log retrieval API (over HTTP).

  Generally, the log files should be treated with the same level of care for persistence *and confidentiality* as the data files: they are absolutely crucial to preserve durably, just like the crdb store data directories, and that's specificially the reason why crdb choses a `logs` sub-directory of the store directory by default. If the customer cares about at-rest confidentiality, the log files should be encrypted together with the data files.

  Right now, we are lucky that the flag `--logtostderr` merely *copies* log data to stderr but still preserves crdb's default behavior, which is to also create discrete files in the `logs` directory. The [statefulset code](https://github.com/cockroachdb/cockroach-operator/blob/master/pkg/resource/statefulset.go#L340), either intendedly or not, was not implemented to explicitly disable writing logging data to files in the store directory, and that's a good thing!

----

**Bottom line**: 

- consider the log files as application data. Keep them on disk.   If you don't, things will BREAK badly (log view via admin UI, retrieval via `cockroach debug zip`).
-  Don't log anything on the stderr. It's misleading and will cause errors down the line when tools get mistakenly developed to parse it. If REALLY anxious about seeing something log just the OPS channel.  The messages remain copied to disk in any case.
- If you want to control disk space usage by logs, use the configuration of a retention window.

udnay:
> consider the log files as application data. Keep them on disk. If you don't, things will BREAK badly (log view via admin UI, retrieval via cockroach debug zip).

This is good information, is there a way to publish logs to both stderr and drop it on disk? We cannot deliver logs in a way that is completely different from what people running on kubernetes expect. The best compromise is likely to `T` the logs. 

>  If REALLY anxious about seeing something log just the OPS channel. 

It's not about being anxious it's that any kubernetes operator expects to be able to use `kubectl logs <my-pod>` and get information. Most logging centralization works on the premise that logs go to stderr and can be collected by k8s and then forwarded onto whatever. This makes it possible to run most workloads on kubernetes and not need to modify your logging stack.

If we were to use the fluentd sink (or some off disk sink that's supported), what does that look like wrt to our tooling, does `debug.zip` still work, do we still store logs on disk?

knz:
> It's not about being anxious it's that any kubernetes operator expects to be able to use kubectl logs <my-pod> and get information

They will get sufficient information with the output from the OPS channel.

Look, you are trying to wish a world into existence that does not exist. CockroachDB is built upon Go, which has defective behavior regarding the stderr channel, and this prevents us from doing both adequate logging and satisfy our log redactability guarantees. *There's nothing we can do about this at this time* and you can't just wish something else into existence.

We have to accept that CockroachDB can't log all what an operator wants on stderr, unless we build an execution script that flips logging on its head:

1. redirect the process' stderr to a file
2. in parallel, tail *some* log files to the container's stderr  (not all: there's one, misleadingly named `cockroach-stderr.log`, which contains misbehaved output)

We could do this, but that would need changing the RUN script inside the container. The one we have currently does not behave properly.

> Most logging centralization works on the premise that logs go to stderr and can be collected by k8s and then forwarded onto whatever

IMHO that's not correct. 

1) the logging centralization works on the promise there's a log collector that a component can _push_ logs to
2) our own CC orchestration considers it best practice to not look at the output of `kubectl logs` to do things. Why do you think we should deviate from our own CC best practices?

> If we were to use the fluentd sink (or some off disk sink that's supported), what does that look like wrt to our tooling, does debug.zip still work, do we still store logs on disk?

The suggested deployment is this:

- use default disk retention for the DEV channel, for troubleshooting.
- copy security-related channels (e.g. USERS, SESSIONS etc) to a collector that guarantees persistent and non-repudiability for the retention window mandated by the local IT corporate policy on security-sentisive logs, possibly with a copy sent to an IDS
- copy operational-relevant channels (e.g. HEALTH, OPS) to the sysops monitoring target - datadog or whatever.

udnay:
> which has defective behavior regarding the stderr channel,

Can you explain this more? Is it specific to `stderr`? Is this true for `stdout` as well? I'm not familiar enough with go's ins and outs.

> our own CC orchestration considers it best practice to not look at the output of kubectl logs to do things. Why do you think we should deviate from our own CC best practices?

CC is different, we own the entire stack there. We cannot enforce these types of decisions on `self-hosted` deployments, similar to how we can't mandate that you need to be running prometheus or a `fluentd` sidecar to aggregate the logs. I don't know all of the historical reasons for our current setup.

knz:
> Can you explain this more? Is it specific to stderr? Is this true for stdout as well? I'm not familiar enough with go's ins and outs.

The go runtime performs asynchronous unbuffered writes to stderr for spurious information/details in a way that the application (in this case, cockroachdb) cannot control.  This breaks our log record delimiters and violates the invariant that the redaction markers be balanced. For example, it could cause confidential information to leak past an implementation of log redaction.

I do not know about stdout. We don't use that for logging AFAIK.

> We cannot enforce these types of decisions on self-hosted deployments, similar to how we can't mandate that you need to be running prometheus or a fluentd sidecar to aggregate the logs.

I thought our CRL business objectives are pretty clear, that we should not create custom-tailored solutions for self-hosted?

> Also without logs on disk what breaks as part of debug.zip I'm guessing it can't collect the logs but the other metrics like CPU and memory profiles would still be collectable?

Without logs, we are unable to troubleshoot issues, and we would be generally unable to support the deployments. No logs, no support service from CRL.

knz:
Summary of what we talked about in person:

- we have an incentive to provide some kube logs compatibility, to obey the rule of "least surprise" for seasoned k8s users
- however, due to aforementioned technical issues inside crdb, we can't dump _all_ crdb logs through the k8s stderr collector
- what then?  we recognize that operators want operator-relevant data. That's what the OPS and HEALTH channels are for.
- additionally, we observe that the OPS and HEALTH channels do not include application data, and thus a log stream corruption due to the go runtime would not cause a confidentiality breach. At the worst, it would cause a loss of records, but the OPS and HEALTH channels typically stream redundant events, so the loss of _some_ events don't impair troubleshootability. 

  So these two channels in particular are suitable for inclusion on the process' stderr output, for collection by k8s
- hence the proposal to have the default config copy those two channels.

- meanwhile, the default config (without additional changes) _continues_ to copy all data (not just those two channels) to files, as required by our various other mechanisms. We're not changing this, and the fact that data goes to disk is not a change compared to previous versions of crdb / previous versions of our k8s specs.

Long term vision:
- we probably want to move off from log files into a per-AZ fluent collector
- one collector per AZ or region, to avoid cross-region networking for logging
- it would also make it possible for customers to customize where they want their log data copied to (s3, datadog, etc) without restarting crdb nodes.
- design for this will be considered in a later phase.

@udnay let me know if I forgot anything

udnay:
Thanks for the summary @knz. @prafull01 lets do a default configuration for OPS, HEALTH to stderr.

@knz should we tune the size of the logs on disk to be low/conservative or are they all on the order of 10's of MB?

knz:
> they all on the order of 10's of MB?

- it's 10MB per file, 100MB retention per log file groups
- there are 5 (6?) groups by default
- we've not heard from customers this was an issue before

udnay:
We can call it out in the docs, it's probably fine.

prafull01:
@udnay Added the default logging configuration `--log={sinks: {stderr: {channels: [OPS, HEALTH], redact: true}}}` and output of the cockroachdb pods logs as follows:
```
prafulll@Velotios-MacBook-Air cockroach-operator % kubectl -n cockroach-operator-system logs cockroachdb-0                
++ expr 800 / 4
++ expr 800 / 4
+ exec /cockroach/cockroach.sh start --join=cockroachdb-0.cockroachdb.cockroach-operator-system:26258,cockroachdb-1.cockroachdb.cockroach-operator-system:26258,cockroachdb-2.cockroachdb.cockroach-operator-system:26258 --advertise-host=cockroachdb-0.cockroachdb.cockroach-operator-system --certs-dir=/cockroach/cockroach-certs/ --http-port=8080 --sql-addr=:26257 --listen-addr=:26258 '--log={sinks: {stderr: {channels: [OPS, HEALTH], redact: true}}}' --cache 200MiB --max-sql-memory 200MiB
*
* INFO: initial startup completed.
* Node will now attempt to join a running cluster, or wait for `cockroach init`.
* Client connections will be accepted after this completes successfully.
* Check the log file(s) for progress. 
*
CockroachDB node starting at 2021-12-31 06:28:39.627107 +0000 UTC (took 27.3s)
build:               CCL v21.1.12 @ 2021/11/15 16:01:13 (go1.15.14)
webui:               https://cockroachdb-0.cockroachdb.cockroach-operator-system:8080
sql:                 postgresql://root@cockroachdb-0.cockroachdb.cockroach-operator-system:26257?sslcert=%2Fcockroach%2Fcockroach-certs%2Fclient.root.crt&sslkey=%2Fcockroach%2Fcockroach-certs%2Fclient.root.key&sslmode=verify-full&sslrootcert=%2Fcockroach%2Fcockroach-certs%2Fca.crt
RPC client flags:    /cockroach/cockroach <client cmd> --host=cockroachdb-0.cockroachdb.cockroach-operator-system:26258 --certs-dir=/cockroach/cockroach-certs/
logs:                /cockroach/cockroach-data/logs
temp dir:            /cockroach/cockroach-data/cockroach-temp413711436
external I/O path:   /cockroach/cockroach-data/extern
store[0]:            path=/cockroach/cockroach-data
storage engine:      pebble
status:              initialized new cluster
clusterID:           cc1dccf6-2b91-42ce-b4f9-80f6dd683f3d
nodeID:              1
```

udnay:
Can we look at getting this merged?

chrisseto:
Sorry for the late comment on this one but I wonder if this will be the ideal way to support logging configurations in going forward.

In the future as well, we may be able to add the ability to reload logging configurations via `SIGHUP` similar to how certificates can be updated on the fly. Having the logging configuration attached directly to the spec will make it much more difficult to leverage such a change in the future (again, if we actually ask for that to be built into the DB).

Have we considered instead allowing users to specify the name of a configmap that contains the configuration that they want to use? That seems more ergonomic in Kubernetes land, would allow us to leverage reloading changes in the future, and would help de-bloat the ClusterSpec.

knz:
@udnay @chrisseto could you get Reviewable configured for this repository? Code reviews with the github interface are cumbersome.

pkg/actor/validate_version.go
```diff
@@ -59,6 +60,26 @@ func (v *versionChecker) GetActionType() api.ActionType {
...
+			cmd.Stderr = &stderr
+			cErr := cmd.Run()
+			if cErr != nil || stderr.String() != "" {
+				log.Error(cErr, "The cockroachdb logging API is set to value that is not supported by the operator, See the default logging configuration here (https://www.cockroachlabs.com/docs/stable/configure-logs.html#default-logging-configuration) ")
+				return errors.New(stderr.String())
+			} else {
+				log.V(DEBUGLEVEL).Info("Validated the logging config")
+			}
+		} else {
+			err := ValidationError{Err: errors.New(fmt.Sprintf("logging configuration %v not supported", cluster.Spec().LogConfig.LogFile))}
```
knz, line 77: This message should include the text of the original error returned by LoggingConfiguration.

pkg/resource/cluster.go
```diff
@@ -295,6 +298,47 @@ func (cluster Cluster) IsFresh(fetcher Fetcher) (bool, error) {
...
+		logYamlBytes, err := json.Marshal(cluster.Spec().LogConfig.LogFile)
+		if err != nil {
+			return "", err
+		}
+		var out bytes.Buffer
+		err = json.Indent(&out, logYamlBytes, "", " ")
+		if err != nil {
+			return "", err
+		}
+		logYaml = "\"" + out.String() + "\""
```
knz, line 313: Have you tested how this behaves if the YAML configuration contains quoted strings?

pkg/resource/statefulset.go
```diff
@@ -356,13 +356,19 @@ func (b StatefulSetBuilder) dbArgs() []string {
...
 			b.Cluster.DiscoveryServiceName(), b.Cluster.Namespace()),
-		"--logtostderr=INFO",
 		b.Cluster.SecureMode(),
 		"--http-port=" + fmt.Sprint(*b.Spec().HTTPPort),
 		"--sql-addr=:" + fmt.Sprint(*b.Spec().SQLPort),
 		"--listen-addr=:" + fmt.Sprint(*b.Spec().GRPCPort),
 	}
 
+	if b.Cluster.IsLoggingAPIEnabled() {
+		logConfig, _ := b.Cluster.LoggingConfiguration()
```
knz, line 366: are you sure it's not necessary to check the error here?

knz (commented):
I broadly agree with Chris that this functionality is defective-on-arrival since it would require restarting all nodes whenever the config is changed, and we expect customers to want to change their logging config while the cluster is running.

However, maybe this amount of defect is *temporarily* acceptable, since it provides an incremental improvement upon the previous situation. 

What is wisdom here? Should we delay this work until https://github.com/cockroachdb/cockroach/issues/77764 gets resolved? Or use the proposed mechanism here for a bit longer?

I think it's worth adding the mechanism proposed here as a stop-gap *but not document it* -- or even, document that this support is "experimental and will be replaced in a later version" so as to discourage users from using it. The very few customers that heavily depend on this can opt into this new feature while knowing in advance they will need to revisit.

chrisseto:
> However, maybe this amount of defect is temporarily acceptable, since it provides an incremental improvement upon the previous situation.

My only concern here is that we might be changing the operator's public API quite fast. If we change this PR to instead accept the name of a `ConfigMap` we'll have an improvement that's open to further improvements without any changes to user configurations.

> I think it's worth adding the mechanism proposed here as a stop-gap but not document it -- or even, document that this support is "experimental and will be replaced in a later version" so as to discourage users from using it. 

I'm not sure what the urgency of this feature is but I feel that the work to convert it to expecting a user created configmap is quite low. I think @prafull01 is already working on updating it?

@udnay or @knz do either of you have an understanding of how urgently needed this feature is?

pkg/resource/statefulset.go
prafull01: We are already validating the logConfig in validate_version.go so if any error would have caught at that point only, either the yaml is not proper or the `cockroach debug check-log-config` is throwing some error

prafull01:
@chrisseto I am already done with most of the changes. I will push the new changes where the input is configmap name which contains the logging config and it is read from the same configmap and passed to the cockroach start command.

pkg/resource/cluster.go
prafull01: Yes I have verified the changes works with both json and yaml

install/operator.yaml
```diff
@@ -141,6 +141,8 @@ rules:
   - configmaps
   verbs:
   - get
+  - list
```
chrisseto, line 144: Do we actually need these additional permissions? I don't see them used anywhere.

pkg/resource/cluster.go
```diff
@@ -319,6 +320,50 @@ func (cluster Cluster) IsFresh(fetcher Fetcher) (bool, error) {
...
+			ObjectMeta: metav1.ObjectMeta{
+				Name: cluster.Spec().LogConfigMap,
+			},
+		}
+		err := fetcher.Fetch(cm)
+		if err != nil {
+			return "", err
+		}
+
+		for _, val := range cm.Data {
```
chrisseto, line 336: nit: As `cm.Data` is a map type, this will be non-deterministic. If a user misconfigures the configmap, it's possible that the logging configuration will drift across reconciliation loops.

Can we instead require that users use a well known key or set of keys with an order of priority? For example:

```go
if ok, val := cm.Data["logging.yaml"]; ok {
    return `"` + val + `"`
}

if ok, val := cm.Data["logging.json"]; ok {
    return `"` + val + `"`
}

// No well known keys found, default to any empty or default logging configuration.
return "\"{sinks: {stderr: {channels: [OPS, HEALTH], redact: true}}}\""
```

chrisseto (approved):
I was initially envisioning that we'd mount the configmap into the Pod and then just specific the file by its name.

I think this approach will work for the immediate future and it gives us the flexibility to switch to file mounting later on!

pkg/resource/cluster.go
prafull01: Yes we can do that, but then we should add a validation in web hook that if user supplies the logging cm in the crdbcluster spec we check that cm exist that specific keys or not. If not we error out from the webhook itself.  

Also, needs to decide on how many keys we allow, IMO two should be enough one for yaml `logging.yaml` and one for json `logging.json`.

chrisseto: After doing some brooding on the matter, here's what I'm thinking: We only support `logging.yaml` as a single key. If that key isn't present, we fallback to the default configuration. There's also an argument for erroring out if `logging.yaml` isn't present, as this would more closely match the behavior of mounting the configmap into a pod and discovering that the `logging.yaml` file isn't present. Either behavior works for me!

This matches up a bit closer to the flags exposed by cockroach, there's only one flag that accepts YAML. JSON, being a subset of YAML, can successfully be provided through that flag.

It also means we don't have to worry about adding in a webhook. In my experience, most Kubernetes misconfigurations are handled by error loops or fallback behavior rather than validation. It would also be very easy for a user to later change or delete the configmap which means that a webhook couldn't provide that much safety for us.

Let me know what you think!

install/operator.yaml
prafull01: This is the error thrown by controller when the permission is not provided:
```
{"level":"info","ts":1649794189.9426928,"logger":"webhooks","msg":"default","name":"cockroachdb"}
{"level":"info","ts":1649794189.98251,"logger":"controller.CrdbCluster","msg":"Running action with name: VersionCheckerAction","CrdbCluster":"default/cockroachdb","ReconcileId":"xKiq62WjKvTaPT2AcJnDFf"}
{"level":"warn","ts":1649794189.9826143,"logger":"controller.CrdbCluster","msg":"starting to check the logging config provided","CrdbCluster":"default/cockroachdb","ReconcileId":"xKiq62WjKvTaPT2AcJnDFf"}
E0412 20:09:49.990532       1 reflector.go:138] external/io_k8s_client_go/tools/cache/reflector.go:167: Failed to watch *v1.ConfigMap: failed to list *v1.ConfigMap: configmaps is forbidden: User "system:serviceaccount:cockroach-operator-system:cockroach-operator-sa" cannot list resource "configmaps" in API group "" at the cluster scope
E0412 20:09:50.984180       1 reflector.go:138] external/io_k8s_client_go/tools/cache/reflector.go:167: Failed to watch *v1.ConfigMap: failed to list *v1.ConfigMap: configmaps is forbidden: User "system:serviceaccount:cockroach-operator-system:cockroach-operator-sa" cannot list resource "configmaps" in API group "" at the cluster scope
E0412 20:09:52.917644       1 reflector.go:138] external/io_k8s_client_go/tools/cache/reflector.go:167: Failed to watch *v1.ConfigMap: failed to list *v1.ConfigMap: configmaps is forbidden: User "system:serviceaccount:cockroach-operator-system:cockroach-operator-sa" cannot list resource "configmaps" in API group "" at the cluster scope
```

jackchi:
@prafull01 @chrisseto I have an existing cluster and I just tried to change the file-defaults from `crdb-v2` to `json` just for a quick test. But after making the change on existing cluster, I see that `CrdbCluster.spec.logConfigMap` gets erased and doesn't appear after I apply the change. 

Also my logging is still not in JSON format. 

I've updated the operator.yml and crds.yaml and updated operator deployment version. 

Were you able to try this on an existing cluster?

jackchi:
> Added the logging configuration feature for cockroachdb as per the following design [document](https://docs.google.com/document/d/1bcn7FRPH4COwaiJakZrz6pdYk7e8RtrMwHL-SJsk5EM/edit)
> 
> Highlights of the changes:
> 
> * CRD changes for taking the logging configmap input for cockroachdb.
> * The logging configmap should have the logging configs in form of json/yaml. See the following example:
> 
> ```
> apiVersion: v1
> data:
>   logconfig.json: |
>     {
>       "sinks": {
>         "file-groups": {
>           "dev": {
>             "channels": "DEV",
>             "filter": "WARNING"
>           }
>         }
>       }
>     }
> kind: ConfigMap
> metadata:
>   name: logconfig
>   namespace: default
> ```
> 
> Or
> 
> ```
> apiVersion: v1
> data:
>   logconfig.yaml: |
>     sinks:
>       file-groups:
>         dev:
>           channels: DEV
>           filter: WARNING
> kind: ConfigMap
> metadata:
>   name: logcm
>   namespace: default
> ```
> 
> * Users provide the configmap name while creating the `CrdbCluster` CR and provide it in `spec` section as follows:
> 
> ```
> spec:
>   logConfigMap: logconfig
> ```
> 
> * For cockroach `version >= v21.1.0`, the logging API is passed to cockroachdb start. If user doesn't provide logging configmap, by default the logging config is:
> 
> ```
> sinks:
>     stderr:
>         channels: OPS
>         redact: true
> ```
> 
> * For cockroach `version <= v20.x`, the logs are sent to stderr `--logtostderr=INFO`
> * Before starting the cluster, it validates the logging API input is proper or not with `cockroach debug check-log-config --log=<log yaml>`, if the logging input is proper then only we create the cluster.
> * Upgraded the cockroach binary version from `v20.2.5` to `v21.1.0` inside the cockroach operator pod.
> 
> **Checklist**
> 
> * [ ]  I have added these changes to the changelog (or it's not applicable).

Small typo:
```yaml
data:
    logconfig.yaml |-
```

jackchi:
@noguchitoshi You are a savior! I also confirmed that by changing the data to `logging.yaml` I could **actually** configure the logs.

prafull01:
I have updated the PR description as per the latest changes which mentions the logging configmap key should be the `logging.yaml`

PR #829, opened 2021-12-02
Give nightly tests a bit longer to be ready

pseudomuto:
Our nightly tests have been failing. At first, we thought it might just
be flakiness, but the issue appears to be consistent.

After running the tests from my local machine, I was able to reproduce
the issue and found that by adding a pause before saying K8s is ready, I
can reliably get the tests to pass.

Clearly, adding a sleep call is not ideal/desirable. I'm hoping to get
this in so we can stop the bleeding so to speak and then figure out what
the actual issue is.

### DIFF
--- a/pkg/testutil/env/env.go
+++ b/pkg/testutil/env/env.go
@@ -19,6 +19,8 @@ package env
 import (
 	"flag"
 	"fmt"
+	"os"
+
 	api "github.com/cockroachdb/cockroach-operator/apis/v1alpha1"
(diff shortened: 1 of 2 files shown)


pkg/testutil/env/env.go
```diff
@@ -19,6 +19,8 @@ package env
 import (
 	"flag"
 	"fmt"
+	"os"
```
pseudomuto, line 22: Not related, but since I was mucking around in here a bit earlier, I figured I'd just leave this in

davidwding (approved):

PR #830, opened 2021-12-06
Properly mark nodes as decommissioned

pseudomuto:
Currently, nodes are decommissioned by executing `cockroach node
decommission --wait=none` and then waiting for `replicas` to hit zero by
polling `cockroach node status --decommission`.

This leaves nodes in the _decommissioning_ state and never "finalizes"
the decommission by setting their state to decommissioned.

This PR adds a call to `cockroach node decommission` (without the `--wait` flag) once the replicas have
gone down to 0, which will mark the node as decommissioned
appropriately.

**Checklist**


davidwding (approved):
fascinating, thanks for fixing!

chrisseto (approved):
LGTM

PR #831, opened 2021-12-07
Added support for SQL ingress

abhishekdwivedi3060:
* Added support for SQL ingress.
Input to be given:

```
ingress:
    sql:
      ingressClassName: nginx
      annotations:
        key: value
      host: sql.test.com
```
**NOTE**: Changing the SQL ingress host will result in crdb cluster rolling restart due to new node certificate generation for SQL host

SQL client works on TCP, so the ingress controller should be able to handle TCP connection via ingress.

Steps to enable TCP connection via Kubernetes Ingress are given here:
https://kubernetes.github.io/ingress-nginx/user-guide/exposing-tcp-udp-services/

**Checklist**

* [*] I have added these changes to the changelog (or it's not applicable).


apis/v1alpha1/cluster_types.go
```diff
@@ -145,6 +145,9 @@ type CrdbClusterStatus struct {
 	// CrdbContainerImage is the container that will be installed
 	// +operator-sdk:csv:customresourcedefinitions:type=status, displayName="CrdbContainerImage",xDescriptors="urn:alm:descriptor:com.tectonic.ui:hidden"
 	CrdbContainerImage string `json:"crdbcontainerimage,omitempty"`
+	// SQLHost is the host to be used with SQL ingress
+	// +operator-sdk:csv:customresourcedefinitions:type=status, displayName="SQLHost",xDescriptors="urn:alm:descriptor:com.tectonic.ui:hidden"
+	SQLHost string `json:"sqlHost,omitempty"`
```
pseudomuto, line 150: Apologies if I'm missing the obvious, but why do we need this as well as the host value in the Ingress definition?

apis/v1alpha1/cluster_types.go
```diff
@@ -255,7 +258,10 @@ type IngressConfig struct {
 	// +optional
 	UI *Ingress `json:"ui,omitempty"`
 
-	// TODO: Add other ingress config for SQL and GRPC here when implemented
+	// (Optional) SQL port to be exposed via ingress
```
pseudomuto, line 261: nit: this isn't just a port. Maybe something like `(Optional) Ingress options for SQL connections` ?

apis/v1alpha1/webhook.go
```diff
@@ -136,20 +136,24 @@ func (r *CrdbCluster) ValidateDelete() error {
 func (r *CrdbCluster) ValidateIngress() (errors []error) {
 	webhookLog.Info("validate ingress", "name", r.Name)
 
-	if r.Spec.Ingress.UI == nil {
-		errors = append(errors, fmt.Errorf("UI related ingress config missing"))
+	if r.Spec.Ingress.UI == nil && r.Spec.Ingress.SQL == nil {
+		errors = append(errors, fmt.Errorf("atleast one of UI and SQL ingresses must be present"))
```
pseudomuto, line 140: nit: `s/atleast/at least/` `s/and/or/`

pkg/actor/generate_cert.go
```diff
@@ -175,6 +189,9 @@ func (rc *generateCert) generateCA(ctx context.Context, log logr.Logger, cluster
 	// if the secret is ready then don't update the secret
 	// the Actor should have already generated the secret
 	if secret.ReadyCA() {
+		if err := ioutil.WriteFile(rc.CAKey, secret.CAKey(), 0600); err != nil {
```
pseudomuto, line 192: Was this just missing before but needed for something?

CHANGELOG.md
```diff
@@ -144,5 +144,4 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Deleted
 
-* Dead code from multiple packages
-* Dependency on ginko and gomega
+*
```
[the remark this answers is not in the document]
abhishekdwivedi3060: typo, removed by mistake. Reverted back

apis/v1alpha1/cluster_types.go
abhishekdwivedi3060: Done

apis/v1alpha1/webhook.go
abhishekdwivedi3060: Done

apis/v1alpha1/cluster_types.go
abhishekdwivedi3060: We need to detect if the SQL host is changed or not. So to keep track of old SQL host and new SQL host (if updated in CR), we are setting current the current host in status

pkg/actor/generate_cert.go
abhishekdwivedi3060: We need to regenerate the node certificates when SQLhost is added/updated, but to regenerate using `cockroach` CLI, CA key and cert is required which are updated in the secret. So here we are writing those key and cert inside the controller on a temp path for node certificate regeneration

apis/v1alpha1/cluster_types.go
davidwding: Is it possible to pull the host from the Ingress object? We don't store any other analogous information on the status - for instance, instead of storing num CRDB nodes on the status, we pull it directly from the StatefulSet.

If this is not possible, then do we also need to store the host for the UI ingress here?

abhishekdwivedi3060: I followed similar approach of `cluster.Spec().Image.Name` and `cluster.Status().CrdbContainerImage`. Though both contains the same info but it's used to take a decision to force trigger Version checker.

Similarly, spec and status SQL host can change during update and we need some differentiator to trigger certificate regeneration before actually going to  `ExposeIngressAction` action.

This is not applicable for UI Ingress because change in UI ingress host does not trigger any other actor. `ExposeIngressAction` will take care of updating the UI ingress Host

davidwding (approved):

pseudomuto (approved):

CHANGELOG.md
```diff
@@ -39,7 +39,7 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 ## Added
 
 * Support for nodeSelectors
-* Support for Ingress (UI only)
+* Support for UI and SQL Ingress
```
pseudomuto, line 42: This doesn't seem right. We didn't add this in v2.3.0. I think we should add a new entry in the Unreleased section rather than changing this one.

abhishekdwivedi3060: Moved to correct location

PR #834, opened 2021-12-27
Fix reconcile annotations init when it's nil

moonsphere:

during learning code of `reconcileAnnotations`, I think it should `SetAnnotations` when init desired object's annotations.

otherwise following put aa should be useless?

**Checklist**

* [x] it's not applicable to the changelog.


pseudomuto (approved):
Thanks for the PR 👍

PR #840, opened 2022-01-06
Don't store bundle here

pseudomuto:
With the updated release process for OperatorHub we no longer need to
keep the bundle around in this repo, since these will be in their
respective upstream repos.

When generating the bundles, we also need to get the SHA of the image
since certified bundles no longer allow tags (since these can be changed
after the fact). In the current process, the image hasn't been released
yet, so we're unable to generate a valid bundle at this point anyways.

I've updated the code here to skip the bundle when running `make
release/new`. The updated release process will now be:

* `make release/new VERSION=x.y.z` (+ PR->Review->Merge)
* Publish docker image
* Tag the release
* Run `make release/generate-bundle`
* Use dirs in ./bundle (now gitignored) to create upstream PRs for the
  certified and certified-rhmp releases.


Makefile
```diff
@@ -355,10 +355,5 @@ PKG_MAN_OPTS ?= "$(PKG_CHANNELS) $(PKG_DEFAULT_CHANNEL)"
 
 # Build the packagemanifests
 .PHONY: release/generate-bundle
-release/generate-bundle: dev/generate
+release/generate-bundle:
 	bazel run //hack:bundle -- $(RH_BUNDLE_VERSION) $(RH_OPERATOR_IMAGE) $(PKG_MAN_OPTS) $(RH_COCKROACH_DATABASE_IMAGE)
-
-#  Build the OPM bundle
-.PHONY: release/opm-build-bundle
-release/opm-build-bundle: release/generate-bundle
```
pseudomuto, line 363: This wasn't being used anymore, so I just 🔥'd it

davidwding: 👋

davidwding (approved):
can you update this doc too? https://cockroachlabs.atlassian.net/wiki/spaces/KO/pages/1753841712/Cockroach+Operator+Release+Process

PR #848, opened 2022-01-26
Add RELEASE_IMAGE_COCKROACH_<version> to manager env

pseudomuto:
When supplying `cockroachDBVersion` in the spec rather than a specific
image, we check for the image in the manager's env. However, at the
moment that env doesn't include the related images.

This PR addresses that by adding them to the generated deployment
manifest, which in turn is used to create the operator.yaml file in
_/install_.

I've verified that everything still works as expected when generating
the bundle by running `make release/generate-bundle` and ensuring the
image values are what we expect.

**Checklist**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 ## Fixed
 
 * Bundle generation for updated OpenShift marketplace requirements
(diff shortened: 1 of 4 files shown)


pseudomuto:
#847 should resolve the build errors here. I'll rebase after the binaries PR is in

davidwding (approved):

PR #849, opened 2022-01-27
Replace kind with k3d

pseudomuto:
When running some tests I was able to reduce the time they take to run
quite a bit by using k3d rather than kind.

In smaller tests, this was significant, for example the smoketests ran
in less than half the time. On larger tests, it shaves a couple of
minutes off (on the ~10 I ran anyways). Pressumably this is just the
difference in the time it takes to start/stop k3d vs kind.

To make this all work, I had to create a custom kubetest2 deployer for
k3d. It's in _e2e/kubetest2-k3d_ and largely is a copy of the kind
deployer updated to use k3d instead.


davidwding:
is this still something we want? seems really nice to have but i'm not sure how much more work is required.

pseudomuto:
This was just something I was playing around with on some flex time. I don't think we actually need to do this, but figured I'd take it for a spin. It probably wouldn't take too much more work to get it working. Think I should 🔥 or get it working?

davidwding:
If it's not too much more work then IMO it's probably worth it to get working? The time savings seem great and I like that it's gesturing at unifying our testing infra across repos.

pseudomuto:
After a rebase and removing the storageClassName from testing volumes, this is ready for review

pkg/testutil/exec/exec.go
```diff
@@ -82,16 +38,16 @@ func StopGKEKubeTest2(clusterName string, zone string, project string) error {
 		"--ignore-gcp-ssh-key",
 	}
 
-	println("Down(): stopin gke cluster...\n")
+	println("Up(): startin gke cluster...\n")
```
davidwding, line 41: lol 🤠

Issue #850, opened 2022-02-02
Omitting volumeMode in pvc spec causes the operator stuck in reconciliation loop


meridional:
This is discovered in https://github.com/cockroachdb/cockroach-operator/issues/845

This seems to be a bug in the `github.com/banzaicloud/k8s-objectmatcher/patch` library that we use (https://github.com/banzaicloud/k8s-objectmatcher/pull/49).

We should consider forcing this setting to the "Filesystem" value, CRDB can't deal with "Block" volumes directly AFAIK.

There are a couple of ways to do this:
- Reject bad volumeMode specs at the creation webhook and instruct the users to fix it;
- Ignore the field and force the setting to "Filesystem" when submitting the sts object.

The second solution is more user friendly, but it will break existing clusters that already have omitted this setting, because
> updates to statefulset spec for fields other than 'replicas', 'template', and 'updateStrategy' are forbidden"

even when setting `volumeMode` to the already in-effect `Filesystem` value.
We can attempt to detect if the cluster is an existing one when submitting the sts object and only take the action if it is not. This feels very brittle and requires significant changes. Not sure if it is worth the risk just to deal with this issue.

So I propose that we go with the first solution.

PR #851, opened 2022-02-02
Validate pvc.volumeMode to circumvent potential reconcilation loop

meridional:
This is an attempt to fix the issue reported
in https://github.com/cockroachdb/cockroach-operator/issues/850
and https://github.com/cockroachdb/cockroach-operator/issues/845.

Currently, there seems to be a bug in
[k8s-objectmaker](https://github.com/banzaicloud/k8s-objectmatcher/pull/49),
which causes the operator to think that the sts
has changed since creation and get stuck in the reconcilation loop.

To mitigate, we check whether `volumeMode` is set,
and set correctly upon cluster creation.

### DIFF
--- a/apis/v1alpha1/webhook.go
+++ b/apis/v1alpha1/webhook.go
@@ -95,6 +95,10 @@ func (r *CrdbCluster) ValidateCreate() error {
 		errors = append(errors, err)
 	}
 
+	if err := r.ValidateVolumeMode(); err != nil {
+		errors = append(errors, err)
+	}
+
 	if len(errors) != 0 {
 		return kerrors.NewAggregate(errors)
 	}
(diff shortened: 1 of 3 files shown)


apis/v1alpha1/webhook.go
```diff
@@ -161,3 +165,19 @@ func (r *CrdbCluster) ValidateCockroachVersion() error {
...
+	claim := r.Spec.DataStore.VolumeClaim
+	if claim == nil {
+		return nil
+	}
+	if claim.PersistentVolumeClaimSpec.VolumeMode == nil {
+		return fmt.Errorf("you have not provided pvc.volumeMode value.")
+	}
+	if *claim.PersistentVolumeClaimSpec.VolumeMode != v1.PersistentVolumeFilesystem {
+		return fmt.Errorf(
+			"you have not provided unsupported pvc.volumeMode, currently only Filesystem is supported.")
```
davidwding, line 180: correction: rm "not"; should be "you have provided unsupported..."

meridional: Thanks for the catch. I can't English... I fixed it in an update.

davidwding (approved):
thank you for fixing this!

meridional:
TFTR!

PR #861, opened 2022-03-07
Change how development defaults are handled

### LINKED ISSUE #390: Change how we handle developer defaults in the makefile and with bazel --stamp
To enable more developers to easily contribute to the development of the operator.

pseudomuto:
Fixes #390 

Currently, we have default values for stamped values defined in multiple
places (some in Makefile, and some in
hack/build/print-workspace-status.sh). There are also quite a few unused
values that we should probably just remove to avoid any confusion.

I've updated the workspace status and Makefile here accordingly.
Defaults are now specified in the Makefile only. If any developer want's
to change them they can supply overrides on the CLI or use a `.envrc`
file with something like [direnv](https://direnv.net/). This file is
already gitignored.


davidwding (approved):
✨

PR #864, opened 2022-03-10
adding in a annotation for the public service

mzupan:
This allows annotations to be set on the public service for the cluster the operator creates. This is needed for things like datadog-agent which can scrape a prometheus endpoint based on an annotation on the service. 

The other use case is running crdb in a peered AWS account where you want to ingress into it. So we can add annotations to the public service to create a NLB in aws

Comments in a few of the code locations for not really sure the best way to do things in go

**Checklist**


pkg/resource/public_service.go
```diff
@@ -48,7 +48,15 @@ func (b PublicServiceBuilder) Build(obj client.Object) error {
...
 	}
 
-	service.Annotations = b.Spec().AdditionalAnnotations
+	annotations :=  make(map[string]string)
+	for k,v := range  b.Spec().AdditionalAnnotations {
+		annotations[k] = v
+	}
+	for k,v := range  b.Spec().PublicServiceAnnotations {
+		annotations[k] = v
+	}
```
mzupan, line 57: the two for loops.. probably not the best way to handle it. I tried a few other ways but if AdditionalAnnotations wasn't set it kept erroring out for a nil map. Open to ideas on how to make this better

PR #865, opened 2022-03-10
Fetch image digests when running release/gen-templates

pseudomuto:
Currently, we have some placeholders in our templates which produce a
token value for each version in crdb-versions.yaml. These tokens are
then replaced with actual image values when we generate the OLM
bundle.

Recently, `docker manifest inspect`, used to fetch information about
remote images, stopped working with the RedHat connect registry. I went
down a few rabbit holes before realizing that all of the info we need is
available from the same API call we use to generate crdb-versions.yaml.

So, rather than fetching the SHAs when we generate the bundle, I figured
we could just have them all written from the templates and skip the
placeholder tokens altogether. Which is what I've done here.

When `make release/gen-templates` is run we fetch all tags from the API
and use the `docker_image_digest` field to also populate the
`redhatImage` field. This means our generated files already have the
pinned versions of the images and that the generate bundle code can be
simplified quite a bit.

I used this to generate the bundle for the v2.5.1 marketplace operator.
I then spun up an OpenShift cluster and made sure it works as expected.

**Checklist**


chrisseto (approved):
LGTM!

PR #866, opened 2022-03-10
Use bazelisk in CI to avoid hard coded bazel versions

pseudomuto:
Rather than upgrading all of the `4.0.0` values in our workflows, I
think we should leverage [bazelisk] for this. This will ensure we've
always got the latest official release being used in CI.

If we ever wanted to fix a version we can always use the env or rcfile
options supported by bazelisk (not used here).

**Checklist**

[bazelisk]: https://github.com/bazelbuild/bazelisk


chrisseto (approved):
LGTM!

PR #868, opened 2022-03-14
Enable support for leader election.

### LINKED ISSUE #863: Manager options missing LeaderElectionID
The operator currently hard crashes when `-enable-leader-election` is specified since the manager api no longer allows using the default leader election id. Because of this issue, running multiple replicas of the operator with leader election cannot be done. This should be as easy as setting the leader election id to a standard value.

pseudomuto:
As mentioned in #863, there is no longer a default value for
`LeaderElectionID` on the manager options. This means that when
`--enable-leader-election` is supplied, the operator fails to start.

I've supplied a default value for the flag here, since most cases won't
need to change this. However, if you intend on running multiple versions
of the operator, having the CLI flag makes it easy to avoid name
collisions.

Fixes #863

**Checklist**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 # [Unreleased](https://github.com/cockroachdb/cockroach-operator/compare/v2.5.1...master)
 
+## Fixed
+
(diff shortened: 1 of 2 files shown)


meridional (approved):

pseudomuto:
TFTR!

PR #870, opened 2022-03-15
Increase timeout for cluster creation in E2E tests

pseudomuto:
Our e2e tests have been failing. From looking at the logs, there are a
number of things that could be the cause.

The first, is that we're seeing cluster creation timeouts. To mitigate
this I've added a shared variable we can use to set the timeout as
necessary. For now, I've bumped it from 500s to 10m (600s).

The second issues is that it seems our openshift kubetest2 thingy
doesn't clean up properly once something fails. This means we need to go
into the Google console and delete some DNS record sets, NAT tables, and
load balancers. This should be updated, but isn't addressed here.


e2e/create/create_test.go
```diff
@@ -30,26 +29,18 @@ import (
...
-// deployed in the cluster helps
-// We may have a threadsafe problem where one test starts messing with another test
-var parallel = *flag.Bool("parallel", false, "run tests in parallel")
-
-var validImage = "cockroachdb/cockroach:v21.1.7"
-var nonExistentImage = "cockroachdb/cockroach-non-existent:v21.1.999"
-var invalidImage = "nginx:latest"
-var crdbVersion = "v21.1.7"
-var relatedImageEnvName = "RELATED_IMAGE_COCKROACH_v21_1_7"
+var (
```
davidwding, line 32: worth plopping these into e2e.go as well? or meh

davidwding (approved):

e2e/create/create_test.go
pseudomuto: I left these here since they weren't used anywhere else. I'm on the meh side 😄

PR #876, opened 2022-04-07
fix NPE error when accessing spec.Image

meridional:
This patch fixes several accesses to
`spec.Image` fields that did not check
for potential `nil` value.

This bug eluded e2e tests because
the `Image` field was always set to
non-nil. A knob to the test cluster
builder is added so that we can properly
test.

### DIFF
--- a/e2e/create/create_test.go
+++ b/e2e/create/create_test.go
@@ -259,7 +259,7 @@ func TestCreateSecureClusterWithCRDBVersionSet(t *testing.T) {
 
 				builder := testutil.NewBuilder("crdb").WithNodeCount(3).WithTLS().
 					WithPVDataStore("1Gi", "standard" /* default storage class in KIND */).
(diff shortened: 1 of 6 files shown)


e2e/create/create_test.go
```diff
@@ -259,7 +259,7 @@ func TestCreateSecureClusterWithCRDBVersionSet(t *testing.T) {
 
 				builder := testutil.NewBuilder("crdb").WithNodeCount(3).WithTLS().
 					WithPVDataStore("1Gi", "standard" /* default storage class in KIND */).
-					WithCockroachDBVersion(crdbVersion)
+					WithCockroachDBVersion(crdbVersion).WithImageObject(nil)
```
pseudomuto, line 262: 👍

pseudomuto (approved):
This LGTM

meridional:
TFTR!

PR #886, opened 2022-04-21
Prefer user added --join flags over default when explicitly passed

neurodrone:
We have been observing a slight issue with the currently default behavior of automatically including the `--join` flag for all the nodes within the StatefulSet. For most cases this works great, however, when we want to make the pods join a different set of nodes and we explicitly pass in the `--join` flag as a part of `additionalArgs` field, we have observed that the pods of the new cluster will still end up happy forming their own quorum even if they are temporarily not able to reach the nodes specified explicitly in the `--join` command that the we passed.

Ideally we'd like it to respect and prefer the `--join` command passed within the `additionalArgs` over the default. If there are reachability issues with the existing nodes then I think it's okay for those to be surfaced explicitly so the folks managing cluster operations can address that issue and have the nodes reconcile after that.

This is a small proposed change to address the above issue. This change is made in backwards compatible way by avoiding to add any new fields to the CRD but I will leave it to you folks to decide if we want any additional fields or not. If people using the Operator do not want to join the new nodes to nodes from an existing cluster they can avoid passing this flag and the hope is the default behavior should work as expected.

### DIFF
--- a/pkg/resource/statefulset.go
+++ b/pkg/resource/statefulset.go
@@ -359,7 +359,6 @@ func (b StatefulSetBuilder) dbArgs() []string {
 	aa := []string{
 		"/cockroach/cockroach.sh",
 		"start",
-		"--join=" + b.joinStr(),
 		fmt.Sprintf("--advertise-host=$(POD_NAME).%s.%s",
 			b.Cluster.DiscoveryServiceName(), b.Cluster.Namespace()),
 		b.Cluster.SecureMode(),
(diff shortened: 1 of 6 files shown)


ADustyOldMuffin:
Fixes #887

neurodrone:
Hi @chrislovecnm, would appreciate if you can 👀 this!  (based on convo with @udnay)

chrisseto (approved):
@ADustyOldMuffin I think this PR needs a rebase to pass the final check.

Otherwise, LGTM.

@neurodrone, I think you might have gotten the wrong "chris". Sorry for the confusion! 😅

neurodrone:
@chrisseto Sorry about that! :) I will do a rebase and push.

neurodrone:
@chrisseto I think I might need an approval to to run the workflows.

chrisseto:
@neurodrone and @ADustyOldMuffin Thank you for the patience on this one!

PR #900, opened 2022-05-19
Increase TerminationGracePeriod to 5m

pseudomuto:
Currently we have the termination grace period set to 1m. This seems a
bit low since we need to do a number of things (e.g. drain connections)
before properly shutting down and it could take a bit longer on busy clusters.

I've bumped this to 5m to give the nodes a little more time. This
represents an upperbound of the time, so should have no speed/perf
penalty in most cases.

**Checklist**


chrisseto (approved):
LGTM

Issue #906, opened 2022-05-26
Operator specifies incorrect service DNS when running in a separate namespace


neurodrone:
## Bug Description

We have observed that the Operator attempts to establish a DB connection to the Cockroach cluster running within the same kubernetes cluster, during one of its state transitions, and succeeds in doing so in all cases but one, where:
* Operator is running in a separate namespace than the `CrdbCluster` custom-resource.

We see the following error when that happens:
```
   message: 'failed to create database connection: opening a DB connection failed
      testing db connection failed: lookup cockroachdb-public on x.x.x.x:53:
      no such host'
    status: Failed
    type: PartitionedUpdate
```

It is attempting to look up DNS record for `cockroachdb-public` service which doesn't exist in its namespace. However, if it did lookup `cockroachdb-public.<namespace>` it would have success finding it.

## To Reproduce

Steps to reproduce:
1. Deploy Operator in a namespace. (e.g. `cockroach-operator-system`)
2. Deploy a `CrdbCluster` in a separate namespace (e.g. `cockroach-cluster`). Make sure that `cockroach-cluster` namespace is passed as an input in the `WATCH_NAMESPACE` environment variable of the Operator's container.
3. A new Cockroach cluster should be spun up as a StatefulSet in the `cockroach-cluster` namespace.
4. Perform an upgrade on the cluster. This can be done by updating the aforementioned `CrdbCluster` custom-resource by either changing its `spec.cockroachDBVersion` field or `spec.image.name`.
5. Observe the above status message show up during the transition phase after the version check completes.

As a result of the above problem, the upgrade never gets initiated on the StatefulSet.

## Expected behavior

No error is seen when conducting `PartitionedUpdate` action and actions like Cockroach server version upgrade complete successfully.

ethan-gallant:
Anyone with this issue who wants to do a quick hackaround you can:

* Create a service in the operator namespace with the same name as your clusters service
* Create an `Endpoints` resource that matches your service name and specify the cockroach actual service IP
```

Examples:
Service
```
apiVersion: v1
kind: Service
metadata:
  name: crdb-public
  namespace: cockroach-operator-system
spec:
  ports:
  - name: grpc
    port: 26258
    protocol: TCP
    targetPort: 26258
  - name: http
    port: 8080
    protocol: TCP
    targetPort: 8080
  - name: sql
    port: 26257
    protocol: TCP
    targetPort: 26257
  type: ClusterIP
```
Endpoints

```
[...]

PR #907, opened 2022-05-26
partitioned_update: use namespaced service name when running in a separate namespace

neurodrone:
In order to decide the service name to reach out for establishing connection, currently, we check whether we run within Kubernetes or outside of it. In order to account for the edge case mentioned in https://github.com/cockroachdb/cockroach-operator/issues/906, when we see the Operator is running within the same Kubernetes cluster, we additionally check if its within the same namespace as the managed clusters or not.

Fixes https://github.com/cockroachdb/cockroach-operator/issues/906.

### DIFF
--- a/pkg/actor/partitioned_update.go
+++ b/pkg/actor/partitioned_update.go
@@ -19,7 +19,7 @@ package actor
 import (
 	"context"
 	"fmt"
-	"github.com/go-logr/logr"
+	"io/ioutil"
 	"os"
 	"strings"
 	"time"
@@ -31,6 +31,7 @@ import (
 	"github.com/cockroachdb/cockroach-operator/pkg/resource"
 	"github.com/cockroachdb/cockroach-operator/pkg/update"
 	"github.com/cockroachdb/errors"
+	"github.com/go-logr/logr"
 	"go.uber.org/zap/zapcore"
 	appsv1 "k8s.io/api/apps/v1"
 	kubetypes "k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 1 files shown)


pkg/actor/partitioned_update.go
```diff
@@ -128,10 +129,14 @@ func (up *partitionedUpdate) Act(ctx context.Context, cluster *resource.Cluster,
...
-	runningInsideK8s := inK8s("/var/run/secrets/kubernetes.io/serviceaccount/token")
+	var (
+		runningInsideK8s = inK8s("/var/run/secrets/kubernetes.io/serviceaccount/token")
+		k8sNamespace     = readK8Sfile("/var/run/secrets/kubernetes.io/serviceaccount/namespace")
+	)
 
-	serviceName := cluster.PublicServiceName()
-	if runningInsideK8s {
+	var serviceName string
+	if runningInsideK8s && k8sNamespace == cluster.Namespace() {
```
chrisseto, line 138: Instead of handling a special case where we happen to be running in the same namespace, let's just always connect to the namespaced or fully qualified service name?
```suggestion
        // We might need .svc.cluster.local as the suffix here?
        serviceName := fmt.Sprintf("%s.%s", cluster.PublicServiceName(), cluster.Namespace())
	if runningInsideK8s {
```

a5r0n: the fully qualified service can be different from `svc.cluster.local` if u set custom local domain

pseudomuto:
This was addressed via #943. @neurodrone thanks so much for the work here

PR #909, opened 2022-05-26
resource/cluster: prefer custom node-tls or client-tls secrets

neurodrone:
We had been seeing issues where the Operator fails to establish DB connection to the cluster when using custom Node and Client TLS certificates. On a closer examination it looked like they were always referencing hardcoded secret names, which were absent when using custom secrets that were passed to the `CrdbCluster` as inputs to `nodeTLSSecret` and `clientTLSSecret`.


davidwding (approved):
Thank you for fixing this and adding a test!

PR #911, opened 2022-05-31
Fix cluster status update logic

davidwding:
Previously, several actors would cancel their context when successful.
This was a relic of the way the reconciliation loop was previously
structured, but now leads to a bug where cluster status is not updated
upon successful action. To address this, this patch deletes the
CancelLoop function entirely.

Addresses https://cockroachdb.slack.com/archives/C01H20CU0A2/p1653707605246189

**Checklist**


davidwding (commented):
> Is it safe to remove the `CancelLoop` function entirely? If not, can we update the docs on it to recommend when it is or isn't safe to use?

After thinking about it I think you're right. I've removed the function entirely.

chrisseto (approved):
LGTM!

PR #913, opened 2022-06-07
actor: correctly detect failed version checker Pods

chrisseto:
Previously, the version checker action could silently fail by if
    given a non-cockroach container. It would incorrectly set the
    cluster version to the error output of the container resulting in
    error such as:
    > failed to reconcile crdb: Service \"crdb\" is invalid:
    > [metadata.labels: Invalid value: \"/bin/bash: line 1:
    > /cockroach/cockroach.sh: No such file or directory\"

    This commit ensures that we the Pod that logs get pulled from has
    exited with a zero code before attempting to parse the version.

**Checklist**


davidwding (approved):
LGTM!

davidwding (approved):
thanks for additional cleanup! ✨

PR #915, opened 2022-06-13
controller: consolidate status update logic

chrisseto:
Previously, Status updates could fail due to version conflicts. This
    commit pulls all status update logic into a helper function that
    utilizes retry.RetryOnConflict to ensure that .Status updates get
    persisted in spite of resource version changes.
    Additionally, this commit removes an early bailout if the controller
    detects a divergence in the currently loaded resource and the API
    version. This change should have no notable affects as it took place
    after any actors ran.
    It also fixes a buglet that overwrote the reconcilers context,
    negating the maximum runtime for a single reconcilation loop.

**Checklist**


meridional (approved):
LGTM

Issue #918, opened 2022-06-16
The program crashes if the image name does not contain colon


hoyhbx:
**What version of operator are you using?**
commit 561cf47d783c368fd8795acb82a5a39099a35984 (HEAD -> master)

**What operating system and processor architecture are you using (`kubectl version`)?**
Ubuntu. 20.04

<details><summary><code>kubectl version</code> Output</summary><br><pre>
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"24", GitVersion:"v1.24.0", GitCommit:"4ce5a8954017644c5420bae81d72b09b735c21f0", GitTreeState:"clean", BuildDate:"2022-05-03T13:46:05Z", GoVersion:"go1.18.1", Compiler:"gc", Platform:"darwin/amd64"}
Kustomize Version: v4.5.4
Server Version: version.Info{Major:"1", Minor:"22", GitVersion:"v1.22.9", GitCommit:"6df4433e288edc9c40c2e344eb336f63fad45cd2", GitTreeState:"clean", BuildDate:"2022-05-19T19:53:08Z", GoVersion:"go1.16.15", Compiler:"gc", Platform:"linux/amd64"}
WARNING: version difference between client (1.24) and server (1.22) exceeds the supported minor version skew of +/-1
</pre></details>

**What did you do?**

We found that crdb-operator will crash if we do not supply version for cockroachdb image. We supplied `cockroachdb/cockroach` for the `spec.image` field, and the operator crashes at [this line](https://github.com/cockroachdb/cockroach-operator/blob/dda0a6308b0f1c64d41e0851a5ef8cb60cd98a37/pkg/resource/cluster.go#L350) 

### Reproduce

We first applied the operator yaml file and crd yaml file to deploy the operator. 
Then we applied the following cr.yaml file to deploy the cockroachdb cluster:
by using `kubectl apply -f cr.yaml -n cockroach-operator-system`

**cr.yaml:**
```yaml
apiVersion: crdb.cockroachlabs.com/v1alpha1
kind: CrdbCluster
metadata:
  name: test-cluster
spec:
  additionalLabels:
    crdb: is-cool
  dataStore:
    pvc:
      spec:
        accessModes:
        - ReadWriteOnce
        resources:
          requests:
            storage: 1Gi
        volumeMode: Filesystem
  image:
    name: cockroachdb/cockroach
  nodes: 3
  resources:
    limits:
      cpu: 2
      memory: 2Gi
    requests:
[...]

PR #922, opened 2022-06-22
Sanity check if the image name has a colon

hoyhbx:
fix https://github.com/cockroachdb/cockroach-operator/issues/918

We fixed the issue by adding sanity check to examine whether a colon is contained in the image name in `IsLoggingAPIEnabled` function. If there is no colon in image name, we will not retrieve index 1 and will directly return false.

**Checklist**

### DIFF
--- a/pkg/resource/cluster.go
+++ b/pkg/resource/cluster.go
@@ -347,7 +347,12 @@ func (cluster Cluster) IsLoggingAPIEnabled() bool {
 	if cluster.Spec().CockroachDBVersion != "" {
 		version = cluster.Spec().CockroachDBVersion
 	} else if cluster.Spec().Image != nil && cluster.Spec().Image.Name != "" {
-		version = strings.Split(cluster.Spec().Image.Name, ":")[1]
+		split_result := strings.Split(cluster.Spec().Image.Name, ":")
+		if len(split_result) > 1 {
+			version = split_result[1]
+		} else {
+			return false
+		}
 	} else {
 		return false
 	}


pkg/resource/cluster.go
```diff
@@ -347,7 +347,12 @@ func (cluster Cluster) IsLoggingAPIEnabled() bool {
 	if cluster.Spec().CockroachDBVersion != "" {
 		version = cluster.Spec().CockroachDBVersion
 	} else if cluster.Spec().Image != nil && cluster.Spec().Image.Name != "" {
-		version = strings.Split(cluster.Spec().Image.Name, ":")[1]
+		split_result := strings.Split(cluster.Spec().Image.Name, ":")
+		if len(split_result) > 1 {
+			version = split_result[1]
+		} else {
+			return false
```
pseudomuto, line 354: Should we add a log line here or something? I'm just thinking that if this condition is hit (no semi-colon in the image name), how would the end user know?

pseudomuto (commented):
Overall this LGTM, thanks for addressing this 👍 

Would you mind adding a test for `IsLoggingEnabled` here to avoid any regressions in the future?

hoyhbx:
Thanks for your response. We have added a unit test in our latest commit.

pkg/resource/cluster.go
hoyhbx: We have tried to add the log line before, but it seems that a parameter that defines the logger needs to be passed to this function. However, this parameter is currently not available in this function.

pseudomuto (commented):
Thanks for the updates. Just a few comments below. This is looking good.

pkg/resource/cluster_test.go
```diff
@@ -85,3 +85,29 @@ func TestClusterTLSSecrets(t *testing.T) {
...
+	for _, tt := range []struct {
+		name      string
+		cluster   *resource.Cluster
+		imageName string
+	}{
+		{
+			name:      "verify image name without colon",
+			cluster:   clusterBuilder.WithImage(customImageName).Cluster(),
+			imageName: customImageName,
+		},
```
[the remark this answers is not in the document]
hoyhbx: Yep, I have added it.

pkg/resource/cluster.go
hoyhbx: Okay, I have added it in the latest commit.

hoyhbx:
@pseudomuto Could you review this so that it can be merged? I think this is a pretty straight-forward fix

PR #930, opened 2022-09-30
Support enterprise encryption

bsnyder788:
This PR adds the ability to enable the enterprise encryption at rest feature.  It supports providing a secret name where the store
encryption key (and the previous one if applicable) is located.

*Checklist*
[ X] I have added these changes to the changelog (or it's not applicable).


udnay:
Hi @bsnyder788, sorry for the delay. We've been figuring out how we wanted to enable this feature internally and started discussing your design here and how we have rolled it out. I dropped the ball on closing the loop. I'm getting the details on our implementation and will provide a review early next week.

PR #943, opened 2022-12-15
[#906] Use FQDN for service address when updating/decommissioning

pseudomuto:
When running the cluster and operator in separate namespaces there's an issue when attempting to connect to the database (from the operator). This is specific to decommissioning and running a partitioned update. As is the code tries to connect using the public service name (`<cr-name>-public`).

To avoid the connection issues described in #906, I've updated this to use the FQDN of the service which is in the following format:

`<cr-name>-public.<namespace>.<cluster-domain>`.

This should make it possible to connect to the database regardless of which namespace it's in.

**Checklist**


chrisseto (approved):
LGTM!

Issue #938, opened 2022-11-14
RH 4.10 and 4.11 Certification


udnay:
The OpenShift Operator certification is the foundation of our Partner Alliance with Red Hat. Maintaining the certification on current versions is a qualification to continue to be a Red Hat Strategic Partner. 

The [certified operator](https://catalog.redhat.com/software/operators/detail/5e9872712989e6a90307acd6) which is available on OpenShift’s OperatorHub has always been the most seamless way to deploy CockroachDB on OpenShift. We need to continue to maintain our certification with the latest versions of OpenShift so prospective users/customers continue to have the deployment option available on newer releases of OpenShift.

Certification of the operator is a self service process on Red Hat’s partner portal. Steps for getting the Operator certified (https://redhat-connect.gitbook.io/partner-guide-for-red-hat-openshift-and-container/):

![certified_operator](https://access.redhat.com/webassets/avalon/d/Red_Hat_Software_Certification-8.53-Red_Hat_Software_Certification_Workflow_Guide-en-US/images/b44dc02fb8b03481bdcfb2e342b70e05/operatorworkflow.png)

The CockroachDB K8s operator is currently certified up to OpenShift v4.9 which was GA on October 18, 2021 and its maintenance support ends on April 18, 2023. Prospective users/customers will care about our certification on OpenShift v4.10 and onwards due to the Red Hat OpenShift Container Platform Life Cycle Policy (https://access.redhat.com/support/policy/updates/openshift)

udnay:
A draft PR of the work can be seen #916. The release process was updated and required an exemption to allow our container to run in a privileged mode.

PR #964, opened 2023-03-02
Update Openshift Release Process

prafull01:
**Install preflight binary on Linux machines**
```
Adding preflight to the list of OpenShift binaries installed. For now,
this is only available on Linux machines and won't be added to anyone
using Darwin.
```

**Publish operator and bundle images for RedHat/OpenShift**
```
Updating the build/release scripts to handle publishing the bundle
images and running preflight on them.

The publish-openshift script downloads the tagged release from
DockerHub, retags it correctly for the RedHat scanning repo, pushes it,
and finally runs the preflight check on it.

The openshift-bundle script largely does the same, only it publishes the
bundle image rather than the operator image. The same script is used for
both the certified and marketplace bundles. Passing `MARKETPLACE=1` will
use the marketplace bundle values rather than the certified ones
(default).
```

**Support Openshift version 4.10 and 4.11 as we have support for k8s 1.25**

Fixes: #938 

**Checklist**

### DIFF
--- a/build/release/teamcity-publish-openshift-bundle.sh
+++ b/build/release/teamcity-publish-openshift-bundle.sh
@@ -17,38 +17,54 @@ set -euxo pipefail
 
 source "$(dirname "${0}")/teamcity-support.sh"
 
+# Default values are defined for the certified bundle.
+RH_PROJECT="5f5a433f9d6546ed7aa8634d"
+RH_REGISTRY="scan.connect.redhat.com"
+RH_REPO="ospid-857fe786-3eb7-4508-aafd-cc74c1b1dc24/cockroachdb-operator-bundle"
+BUNDLE_DIR="bundle/cockroachdb-certified"
 
-tc_start_block "Variable Setup"
-VERSION="v"$(cat version.txt)
-# Matching the version name regex from within the cockroach code except
-# for the `metadata` part at the end because Docker tags don't support
-# `+` in the tag name.
-# https://github.com/cockroachdb/cockroach/blob/4c6864b44b9044874488cfedee3a31e6b23a6790/pkg/util/version/version.go#L75
-image_tag="$(echo "${VERSION}" | grep -E -o '^v(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)(-[-.0-9A-Za-z]+)?$')"
-#                                         ^major           ^minor           ^patch         ^preRelease
-
-if [[ -z "$image_tag" ]] ; then
-    echo "Invalid VERSION \"${VERSION}\". Must be of the format \"vMAJOR.MINOR.PATCH(-PRERELEASE)?\"."
-    exit 1
+# If this is the marketplace bundle, update accordingly.
+if ! [[ -z "${MARKETPLACE}" ]]; then
+  RH_PROJECT="61765afbdd607bfc82e643b8"
+  RH_REPO="ospid-61765afbdd607bfc82e643b8/cockroachdb-operator-bundle-marketplace"
+  BUNDLE_DIR="bundle/cockroachdb-certified-rhmp"
 fi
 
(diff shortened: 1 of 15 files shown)


build/release/teamcity-publish-openshift-bundle.sh
```diff
@@ -17,38 +17,54 @@ set -euxo pipefail
...
+  tc_end_block "Make and push bundle image"
+}
 
-# TODO(rail): switch to bazel generated images when it supports "FROM: scratch"
-cd deploy/certified-metadata-bundle/cockroach-operator
-docker build -t $image -f bundle.Dockerfile .
-docker push $image
+run_preflight() {
+  bazel build //hack/bin:preflight
+  PFLT_PYXIS_API_TOKEN="${REDHAT_API_TOKEN}" bazel-bin/hack/bin/preflight \
```
pseudomuto, line 66: Is this token already in place in TC?

hack/bundle.sh
```diff
@@ -84,11 +84,17 @@ generate_bundle() {
...
   adapt_csv "${dir}" "${img}"
+
+  # move the dockerfile into the bundle directory and make it valid
+  sed \
+    -e "/\s*tests\/scorecard/d" \
+    -e "s+${dir}/++g" \
+    bundle.Dockerfile > "${dir}/Dockerfile"
+
+  rm bundle.Dockerfile
+	rm "${dir}/manifests/cockroach-operator-webhook-service_v1_service.yaml"
```
pseudomuto, line 97: /nit formatting

hack/redhat.sh
```diff
@@ -0,0 +1,118 @@
...
+  # into the build directory, we'll have access to _all_ the files.
+  if [[ -n "${BUILD_WORKSPACE_DIRECTORY}" ]]; then
+    cd "${BUILD_WORKSPACE_DIRECTORY}"
+  fi
+
+  case "${1:-}" in
+    operator)
+			publish_operator_image
+      preflight_operator;;
+    bundle)
```
pseudomuto, line 48: /nit again, formatting. Typically 2 space indent for bash files. Bonus points if the whole file uses the same 😄

pseudomuto (commented):
A few nits, but this is looking good. I'm assuming you've tried this with the --dry-run option and found no issues building/publishing the images?

build/release/teamcity-publish-openshift-bundle.sh
prafull01: I am not sure as I don't have access to TC. Can you please check this token is added or not?

hack/bundle.sh
prafull01: Done

hack/redhat.sh
prafull01: Done

prafull01 (commented):
Yes. I have build the images, pushed to personal repo and tested those images. This is the way I have done testing.

build/release/teamcity-publish-openshift-bundle.sh
prafull01: I am not sure. I don't have access to TC. Can you please check in the TC if we have added this token

hack/bundle.sh
prafull01: Done

hack/redhat.sh
prafull01: Done

prafull01:
Tests is passing locally. Logs points to the following issue in GitHub actions https://github.com/actions/runner/issues/2468 and has been seen by many people recently

pseudomuto (approved):

build/release/teamcity-publish-openshift-bundle.sh
rail: `REDHAT_API_TOKEN` is set in TeamCity, correct.

PR #976, opened 2023-04-17
Fix the clusterStatus should reflect correct state of cluster

### LINKED ISSUE #951: When creating a cluster the clusterstatus indicates 'failed', but the cluster successfully started.
## Summary
**Partner Involved:** SpectroCloud 
Operator cluster status is showing the cluster as failed, but the cluster is online and available. 
 I have a partner that is trying to programmatically check the state of a  cluster created by our operator.

When running: `kubectl get CrdbCluster -o json`   to review the '`status`' and '`condition` '
 of the deployed CRDB cluster, the `clusterStatus` is listed as "Failed".  However, the cluster is available. I can connect and run DDL,DML statements.   The partner is seeing the same behavior. 

## Steps to reproduce
**My actions/env:**
1. Start Minikube
2. Install the operator  [docs](https://www.cockroachlabs.com/docs/stable/deploy-cockroachdb-with-kubernetes.html#install-the-operator).
3.  Initialize the cluster following our [docs](https://www.cockroachlabs.com/docs/stable/deploy-cockroachdb-with-kubernetes.html?#initialize-the-cluster).

## Expected Result
The expected result is when running the following command the `clusterStatus` is listed as 'running' and cluster is online and available.   
 Command: `kubectl get CrdbCluster -o=jsonpath='{range .items[*]}{.metadata.name}{"\t"}{"status="}{.status.clusterStatus}{"\n"}{end}'`

## Actual Result
After running the command: `kubectl get CrdbCluster -o=jsonpath='{range .items[*]}{.metadata.name}{"\t"}{"status="}{.status.clusterStatus}{"\n"}{end}'` that the `clusterStatus` is in a `cockroachdb	status=Failed` state but the cluster is available.

Here are the logs: 
[cockroach-operator-manager-7db5d587b7-qlxws.log](https://github.com/cockroachdb/cockroach-operator/files/10415246/cockroach-operator-manager-7db5d587b7-qlxws.log)

prafull01:
Fixes #951 

Currently, we are storing the CR in resource.Cluster object and changing this CR during the course of reconcile and at last call the update function. This causes the `object has been already modified, please apply your changes to the latest version and try again`

With this change, we take the copy of the object before reconcile, run the changes on our CR, then patch will take diff for the before and after of object to find only the changed fields. This patch operation is more safer in terms of updating the CR.

With this change, cluster is in correct state after crdb is provisioned:
Screenshot 2023-04-18 at 3 00 49 AM

### DIFF
--- a/pkg/controller/cluster_controller.go
+++ b/pkg/controller/cluster_controller.go
@@ -118,11 +118,12 @@ func (r *ClusterReconciler) Reconcile(ctx context.Context, req reconcile.Request
 
 	cluster := resource.NewCluster(cr)
 	cluster.Fetcher = fetcher
+	cleanClusterObj := cluster.Unwrap()
 	// on first run we need to save the status and exit to pass Openshift CI
 	// we added a state called Starting for field ClusterStatus to accomplish this
 	if cluster.Status().ClusterStatus == "" {
 		cluster.SetClusterStatusOnFirstReconcile()
-		if err := r.updateClusterStatus(ctx, log, &cluster); err != nil {
+		if err := r.updateClusterStatus(ctx, log, &cluster, cleanClusterObj); err != nil {
 			log.Error(err, "failed to update cluster status")
 			return requeueIfError(err)
 		}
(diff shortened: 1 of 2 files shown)


pseudomuto (approved):
This LGTM 👍

PR #977, opened 2023-04-18
Update the golangci-lint version from 1.50.1 to 1.51.1

prafull01:
golangci-lint version is updated from the 1.50.1 because it is hitting the OOM issue with the Go 1.20 version and its fix has been done in 1.51.1.

### DIFF
--- a/e2e/openshift/openshift_packaging_test.go
+++ b/e2e/openshift/openshift_packaging_test.go
@@ -233,9 +233,10 @@ var letters = []rune("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ")
 
 // randSeq returns a string of n letters
 func randSeq(n int) string {
+	r := rand.New(rand.NewSource(time.Now().UnixNano()))
 	b := make([]rune, n)
 	for i := range b {
-		b[i] = letters[rand.Intn(len(letters))]
+		b[i] = letters[r.Intn(len(letters))]
 	}
 	return string(b)
 }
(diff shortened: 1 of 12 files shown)


e2e/openshift/openshift_packaging_test.go
```diff
@@ -233,17 +233,17 @@ var letters = []rune("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ")
...
-		b[i] = letters[rand.Intn(len(letters))]
+		b[i] = letters[r.Intn(len(letters))]
 	}
 	return string(b)
 }
 
 // wriiteFile writes out b byte array to a random file in the
 // os.TempDir and returns the name of that file.
 func writeFile(t *testing.T, b []byte) string {
-	rand.Seed(time.Now().UnixNano())
```
pseudomuto, line 246: This just wasn't needed?

pseudomuto (approved):
LGTM 👍

e2e/openshift/openshift_packaging_test.go
prafull01: rand.Seed() is deprecated in Go 1.20. This is the new way to generate random strings as per Go 1.20.

Lint is failing saying `rand.Seed()` is deprecated, hence this change is necessary.

prafull01: Lint job failure without this change: https://github.com/cockroachdb/cockroach-operator/actions/runs/4728824923/jobs/8390733554

prafull01: We have moved the alternative function call to `randSeq` function where we generate the random string. Earlier rand.Seed initialise the default resource once for random method. Now you generate the `rand.Rand` with NewSource function in `randSeq` method itself and use that Rand object to generate the random string.

udnay (approved):

PR #981, opened 2023-05-17
Update the CRDB versions

prafull01:
- Update the cockroach operator with latest CRDB versions
- Added support for Mac m1 to run tests locally
- Upgrade test was failing because of 8 parallel process, reduced it to 4 parallel test cases.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -94,7 +94,7 @@ test/e2e-short:
 test/e2e/testrunner-k3d-%: PACKAGE=$*
 test/e2e/testrunner-k3d-%:
 	bazel run //hack/k8s:k8s -- -type k3d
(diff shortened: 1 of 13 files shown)


udnay:
@prafull01 23.1.2 came out today, can we add it in and then publish the operator?

kangzhang:
We tried upgrading our cluster to 23.1.1 without this PR - and it failed on the version check because the resource is too low for the job. See more details https://github.com/cockroachdb/cockroach-operator/issues/983

Wonder if we should bump the resource limits (for vcheck job) as we bump the versions here.

echozio:
> We tried upgrading our cluster to 23.1.1 without this PR - and it failed on the version check because the resource is too low for the job. See more details #983
> 
> Wonder if we should bump the resource limits (for vcheck job) as we bump the versions here.

Also had issues with this, however it did eventually go through after an hour or so.

rail (approved):

PR #984, opened 2023-06-01
Update the bundle.sh to generate the bundle through opm

prafull01:
- Changed the source file name from `opm` to `darwin-amd64-opm` as per change in the tar file.
- Removed deletion of webhook service from bundle which is required in bundle for webhook


pseudomuto (approved):
LGTM 👍

PR #997, opened 2023-07-11
Remove the duplicate cockroachdb versions generated by the API response

prafull01:
Right now, we pull the images from the `https://catalog.redhat.com/` by calling an API. The response from the API is sending tag `v20.1.4` and `v20.1.5` twice and hence it is added to our deployment twice.

I have added a check that any image is used only once even if we receive duplicate response from API.


rail (approved):
LGTM. I wonder why the api return these images twice. I will take a look at that separately.

PR #998, opened 2023-07-21
Fix the jobs to release on openshift certification

prafull01:
- Added new quay.io registry to push image for redhat certification.
- Added a new RELATED_IMAGE_COCKROACH_OPERATOR env variable in CSV needed for redhat certification.
- Made sure the CSV we generate is in sync with it is generated by the operator-sdk

### DIFF
--- a/build/release/teamcity-publish-openshift.sh
+++ b/build/release/teamcity-publish-openshift.sh
@@ -18,8 +18,8 @@ set -euxo pipefail
 source "$(dirname "${0}")/teamcity-support.sh"
 
 RH_PROJECT_ID="5e6027425c5456060d5f6084"
-RH_REGISTRY="scan.connect.redhat.com"
-RH_OPERATOR_IMG="${RH_REGISTRY}/ospid-cf721588-ad8a-4618-938c-5191c5e10ae4/cockroachdb-operator:${TAG}"
+RH_REGISTRY="quay.io"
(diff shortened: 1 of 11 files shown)


build/release/teamcity-publish-openshift.sh
```diff
@@ -18,8 +18,8 @@ set -euxo pipefail
 source "$(dirname "${0}")/teamcity-support.sh"
 
 RH_PROJECT_ID="5e6027425c5456060d5f6084"
-RH_REGISTRY="scan.connect.redhat.com"
-RH_OPERATOR_IMG="${RH_REGISTRY}/ospid-cf721588-ad8a-4618-938c-5191c5e10ae4/cockroachdb-operator:${TAG}"
+RH_REGISTRY="quay.io"
+RH_OPERATOR_IMG="${RH_REGISTRY}/redhat-isv-containers/${RH_PROJECT_ID}:${TAG}"
```
rail, line 22: I think we also need to tweak the docker username, similar to crdb, see https://github.com/cockroachdb/cockroach/blob/d293bb3acd9b7f67ac793ddb6a0836e637864629/build/teamcity/internal/cockroach/release/publish/publish-redhat-release.sh#L21. I can create a login for this.

rail (changes requested):

build/release/teamcity-publish-openshift.sh
rail: Oh, the wrapper function takes care of those. LGTM.

rail (approved):

PR #1024, opened 2024-03-13
Update CSV and metadata annotations for openshift certification

prafull01:
Openshift certification needs some additional annotations which was not generated by `make release/generate-bundle`. I have added logic to generate those annotations so that the generated bundle passes certification in one attempt.


rail (approved):
LGTM, :shipit:

Issue #1056, opened 2024-10-17
Depreciated --port & --http-port are used in statefulset


udnay:
--port & --http-port has been depreciated, but they're still used in operator
It's a risk if we ever remove these parameters.

PR #1058, opened 2024-10-21
Remove --http-port and --port from the cockroach operator

prafull01:
Fixes: #1056


install/crds.yaml
```diff
@@ -1084,12 +1084,12 @@ spec:
                     type: boolean
                 type: object
               grpcPort:
```
udnay, line 1086: Is this still just a port? Do we need to update the CRD here for something like `grpcAddr` and `httpAddr`

pkg/resource/testdata/TestStatefulSetBuilder/automount_sa.golden
```diff
@@ -31,7 +31,7 @@ spec:
         - /bin/bash
         - -ecx
         - 'exec /cockroach/cockroach.sh start --advertise-host=$(POD_NAME).test-cluster.test-ns
-          --certs-dir=/cockroach/cockroach-certs/ --http-port=8080 --sql-addr=:26257
+          --certs-dir=/cockroach/cockroach-certs/ --http-addr=:8080 --sql-addr=:26257
           --listen-addr=:26258 --log="{sinks: {stderr: {channels: [OPS, HEALTH], redact:
```
udnay, line 35: Wait, did we already make the change to `listen-addr` in the operator? What value is it pulling from?

config/crd/bases/crdb.cockroachlabs.com_crdbclusters.yaml
```diff
@@ -1086,12 +1086,12 @@ spec:
                     type: boolean
                 type: object
               grpcPort:
-                description: '(Optional) The database port (`--port` CLI parameter
+                description: '(Optional) The database port (`--listen-addr` CLI parameter
```
udnay, line 1089: Are these changes backwards compatible?

pkg/scale/drainer.go
```diff
@@ -215,7 +215,7 @@ func (d *CockroachNodeDrainer) executeDrainCmd(ctx context.Context, id uint, gRP
 // a node which will transition it from `decommissioning` to `decommissioned`. This should be executed
 // after it's confirmed that there are 0 replicas on the node.
 func (d *CockroachNodeDrainer) markNodeAsDecommissioned(ctx context.Context, id uint, gRPCPort int32) error {
-	cmd := []string{"./cockroach", "node", "decommission", fmt.Sprintf("%d", id), fmt.Sprintf("--port=%d", gRPCPort)}
+	cmd := []string{"./cockroach", "node", "decommission", fmt.Sprintf("%d", id), fmt.Sprintf("--host=:%d", gRPCPort)}
```
udnay, line 218: What does `--host=:<grpcPort>` default to? Is it localhost?

install/crds.yaml
prafull01: If we change the fields from port to `grpcAddr` and http-port to `httpAddr`, we need to have some kind of migration to actually migrate the old CRs to these new values. This is why to keep it simple I have just changed the description and not fields.

pkg/resource/testdata/TestStatefulSetBuilder/automount_sa.golden
prafull01: Yes we already have change of listen-addr which is equal to `:grpcPort`, I did the same thing from http-port to http-addr=`:httpPort`

config/crd/bases/crdb.cockroachlabs.com_crdbclusters.yaml
prafull01: Yes. I have only changed the description and not the field so it is backward compatible.

pkg/scale/drainer.go
prafull01: Description specifies:
```
--host <addr/host>[:<port>]           
                                             CockroachDB node to connect to. This can be specified either as an
                                             address/hostname, or together with a port number as in -s myhost:26257. If
                                             the port number is left unspecified, it defaults to 26257. An IPv6 address
                                             can also be specified with the notation [...], for example [::1]:26257 or
                                             [fe80::f6f2:::]:26257.
                                             Environment variable: COCKROACH_HOST
                                             (default :26257)
```
It is similar to listenAddr or httpAddr where we can only specify port with `:port`

apis/v1alpha1/webhook.go
```diff
@@ -59,16 +59,28 @@ func (r *CrdbCluster) SetupWebhookWithManager(mgr ctrl.Manager) error {
 func (r *CrdbCluster) Default() {
 	webhookLog.Info("default", "name", r.Name)
 
-	if r.Spec.GRPCPort == nil {
-		r.Spec.GRPCPort = &DefaultGRPCPort
+	if r.Spec.GRPCPort == nil && r.Spec.ListenAddr == nil {
```
udnay, line 62: How does this work, I'm not super familiar with it. Do we need to persist the spec somewhere? I know we pull it often from the k8s api, do we need to update it if we change some of the values here?

pkg/resource/cluster.go
```diff
@@ -48,7 +49,6 @@ const (
 
 func NewCluster(original *api.CrdbCluster) Cluster {
```
udnay, line 50: This is what I'm curious about, does the `original` crdb cluster get the defaults or will we always need to do the migration. I guess we'll always do the migration so that there is no conflict between the persisted object and whatever the customer is using to update it.

udnay (approved):
Please make sure that the tests all pass on the branch, other than that LGTM

apis/v1alpha1/webhook.go
prafull11: This Default() func always gets called in mutating configuration so whenever a create or update request comes on this func gets called and the changes we mutate on crdbCluster goes to apiserver and it gets persisted to etcd.

Now when in controller you get the crdbcluster object from apiserver it returns the object with all the mutation we did in Default() func.

udnay: So if I do `kubectl get crdbcluster foo -o yaml` I'll see `listenAddr` set now?

pkg/resource/cluster.go
prafull11: This will always gets the object which is processed by default (see my prev comment). Whenever we get the object in controller we get the object which has already mutated using Default func. so we don't have to migrate it again. This is the reason I have removed the call to Default func again. This could cause migration on non migrated object.

If that call is present, after upgrade the migration doesn't happen on the actual CR, but in our logic that func does the migration which is two different states for the same object. We should always use the state which is persisted in etcd.

prafull11:
> Please make sure that the tests all pass on the branch, other than that LGTM

For tests to pass, I am starting a webhook server with controller manager so that this Default() func gets called and does the basic mutation on crdbcluster object. It will need some changes on test side so I will ask for a review again.

apis/v1alpha1/webhook.go
prafull11: Yes. If you create a crdbcluster CR with these changes you will see the `listenAddr`, `httpAddr` and `sqlAddr` set.

If you have old CR, then you don't see it (so no automatic rolling restart). However as soon as you update that CR with new image, your rolling restart will happen and now you see addresses instead of ports.

prafull01:
> Please make sure that the tests all pass on the branch, other than that LGTM

For tests to pass, I am starting a webhook server with controller manager so that this Default() func gets called and does the basic mutation on crdbcluster object. It will need some changes on test side so I will ask for a review again.

pkg/resource/cluster.go
prafull01: This will always gets the object which is processed by default (see my prev comment). Whenever we get the object in controller we get the object which has already mutated using Default func. so we don't have to migrate it again. This is the reason I have removed the call to Default func again. This could cause migration on non migrated object.

If that call is present, after upgrade the migration doesn't happen on the actual CR, but in our logic that func does the migration which is two different states for the same object. We should always use the state which is persisted in etcd.

apis/v1alpha1/webhook.go
prafull01: Yes. If you create a crdbcluster CR with these changes you will see the listenAddr, httpAddr and sqlAddr set.

If you have old CR, then you don't see it (so no automatic rolling restart). However as soon as you update that CR with new image, your rolling restart will happen and now you see addresses instead of ports.

PR #1070, opened 2024-11-25
vcheck job resource limits should be greater than resource requests

prafull01:
version checker job is getting OOM on openshift sometimes. The resource requests and limits are same for the version checker job which is not ideal. There should be some wiggle room if there is memory spike at some point, hence increased the resource limits for the version checker job.

### DIFF
--- a/pkg/resource/job.go
+++ b/pkg/resource/job.go
@@ -141,7 +141,7 @@ func (b JobBuilder) MakeContainers() []corev1.Container {
 			Resources: corev1.ResourceRequirements{
 				Limits: corev1.ResourceList{
 					corev1.ResourceCPU:    apiresource.MustParse("300m"),
-					corev1.ResourceMemory: apiresource.MustParse("256Mi"),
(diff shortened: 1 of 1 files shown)


pkg/resource/job.go
```diff
@@ -140,8 +140,8 @@ func (b JobBuilder) MakeContainers() []corev1.Container {
 			ImagePullPolicy: b.GetImagePullPolicy(),
 			Resources: corev1.ResourceRequirements{
 				Limits: corev1.ResourceList{
-					corev1.ResourceCPU:    apiresource.MustParse("300m"),
-					corev1.ResourceMemory: apiresource.MustParse("256Mi"),
+					corev1.ResourceCPU:    apiresource.MustParse("500m"),
```
pritesh-lahoti, line 143: Could we defer from updating the CPU limit unless required?

prafull01: Done

pritesh-lahoti (approved):

PR #1074, opened 2024-12-19
Add Support for linux arm64 images

prafull01:
- Add support for linux arm64 images.
- Upgrade cockroach binary from v21.2.0 to v22.2.19 as it has official support of arm64 image.

### DIFF
--- a/hack/bin/deps.bzl
+++ b/hack/bin/deps.bzl
@@ -13,7 +13,6 @@
 # limitations under the License.
 
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_file", "http_archive")
-load("@io_bazel_rules_docker//container:container.bzl", "container_pull")
 load("@bazel_gazelle//:deps.bzl", "go_repository")
 
 # This controls the version for all openshift binaries (opm, oc, opernshift-install, etc.)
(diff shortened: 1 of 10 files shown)


hack/bin/deps.bzl
```diff
@@ -456,28 +455,43 @@ def install_preflight():
 def install_crdb():
     http_archive(
        name = "crdb_darwin", # todo fix or remove
-       sha256 = "bbbd0a75f81d3df4acd139fdc7f0961480161454db24f25263c9276c3959db54",
-       urls = ["https://binaries.cockroachdb.com/cockroach-v21.2.0.darwin-10.9-amd64.tgz"],
+       sha256 = "79fb1669678b802ae891ec3e005efa801c0f970fec4eb6af7ac89cdb6b991b42",
+       urls = ["https://binaries.cockroachdb.com/cockroach-v22.2.19.darwin-10.9-amd64.tgz"],
```
rail, line 459: Maybe use a newer version? This version is already not supported. Maybe use 24.3.1?

prafull01: Reason for being not use the latest version is the cockroach start command used in the statefulset has some arguments which are deprecated. Those arguments needs to be upgraded with backward compatibility to upgrade to latest version.  It lies outside the scope of this ticket

This is the reason I have upgraded to the minimum version where arm64 is supported.

.github/workflows/e2e-kind-create.yaml
```diff
@@ -26,7 +26,7 @@ on:
 
 jobs:
   e2e-kind-create:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-latest
```
pritesh-lahoti, line 29: nit: misplaced space

pritesh-lahoti (approved):

rail (approved):

PR #1077, opened 2025-01-27
Fix the development workflow in make dev/build.

prafull01:
Removed rules_k8s bazel dependency as it is deprecated and not supported with rules_oci


pritesh-lahoti (approved):

NishanthNalluri (approved):
LGTM

PR #1095, opened 2025-04-02
Update the openshift versions to be supported 4.19

prafull01:


pritesh-lahoti:
Thank you for the quick turn-around, @prafull01 !

PR #1121, opened 2025-07-22
Update the base image to ubi9 to resolve moderate vulnerabilities in ubi8 image.

prafull01:


pritesh-lahoti (approved):
Did this help us address the vulnerabilities?
If so, we could update the PR description to capture the same.

PR #1129, opened 2025-12-15
pkg/resource: Add migration label support to CrdbCluster CR

NishanthNalluri:
To support migrating CockroachDB clusters from the public operator to the CockroachDB operator through coexistence, we would need additional logic to prevent the operator from scaling up StatefulSet pods after they have been manually scaled down.

This commit introduces a new migration label(`crdb.io/stop-public-operator-reconcile`) for the CrdbCluster resource. When this label is applied, the reconciler is instructed to skip updates to the relevant StatefulSet replicas. This ensures the operator does not interfere with the controlled scale-down process required during the migration phase.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -373,4 +373,3 @@ publish-operator-openshift:
 .PHONY: release/publish-openshift-bundle
 release/publish-openshift-bundle:
 	./build/release/teamcity-publish-openshift-bundle.sh
-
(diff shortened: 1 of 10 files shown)


pkg/resource/statefulset.go
```diff
@@ -51,6 +51,8 @@ const (
 	DbInitContainerName = "db-init"
 
 	terminationGracePeriodSecs = 300
+
+	migrationLabel = "crdb.io/migrating"
```
pritesh-lahoti: nit: could add a comment explaining the requirement/behavior of this label.

PR #1147, opened 2026-02-27
Add hostAliases support for CrdbCluster pods

eXist-FraGGer:
_`CrdbCluster` did not expose a way to configure pod-level `hostAliases`, so users could not inject custom `/etc/hosts` entries for CockroachDB pods via the operator API._.

**Checklist**

* [x] adds spec.hostAliases to CrdbCluster API
* [x] propagates hostAliases to StatefulSet pod template
* [x] updates deepcopy for the new field
* [x] extends CRD schemas in config/crd/bases and install/crds.yaml
* [x] adds CSV spec descriptor for hostAliases
* [x] updates example templates with hostAliases snippet
* [x] adds StatefulSetBuilder golden test coverage for hostAliases

**Use case:**
Deployed in minikube `istio+cockroach+keycloak` and configure `hosted local domain zone` (e.g. `pdnsutil zone`).
The issue is to setup `oauth2` at cockroach via keycloak (Own OIDC Provider):

You can not use internal address `.svc.cluster.local` because redirect url at browser will not resolve internal kubernetes address.
So you are using `hosted local zone` e.g. `keycloak.dev.local`, but it will be resolved to 127.0.0.1 - this is wrong and this is why we must have `hostAliases` to configure hostname `keycloak.dev.local` to `istio ingress gateway cluster IP`.

### DIFF
--- a/apis/v1alpha1/cluster_types.go
+++ b/apis/v1alpha1/cluster_types.go
@@ -137,6 +137,9 @@ type CrdbClusterSpec struct {
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Map of nodeSelectors to match when scheduling pods on nodes"
 	// +optional
 	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
+	// (Optional) HostAliases is an optional list of hosts and IPs that will be injected into the pod's hosts file if specified.
+	// +optional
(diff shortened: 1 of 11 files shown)


eXist-FraGGer:
@NishanthNalluri Hello, let's review and discuss as it is very critical configuration.

